PR #461, opened 2023-03-06
(Pruning,BFS) does no longer require an upper bound.

pontusmelke:


JPryce-Aklundh (approved):
lgtm 👍 
If this 5.7 specific, however, it should not be merged yet, as dev branch is still 5.6 due to the delay in release.

PR #491, opened 2023-03-29
Add a PDF-specific title

recrwplay:
When the title of the first page in a manual changes from something like 'Neo4j Cypher Manual' we need to add a condition to set a meaningful title for the PDF.


JPryce-Aklundh (approved):
Thanks!

PR #500, opened 2023-04-06
Restructure index section

Hunterness:
This will hopefully remove some confusion for users around just showing off the command vs an example as well as removing examples for index providers when only one provider exists.

It also updates the links to not mention `administration` as the index chapter does not live in the administration chapter.

### DIFF
--- a/modules/ROOT/pages/indexes-for-search-performance.adoc
+++ b/modules/ROOT/pages/indexes-for-search-performance.adoc
@@ -323,74 +323,6 @@ listing indexes require xref::administration/access-control/database-administrat
 xref:query-tuning/using.adoc[Planner hints and the USING keyword] describes how to make the Cypher planner use specific indexes (especially in cases where the planner would not necessarily have used them).
 
 
-[[indexes-single-vs-composite-index]]
-== Composite index limitations
-
-Like single-property range indexes, composite range indexes support all predicates:
-
-* equality check: `n.prop = value`
-* list membership check: `n.prop IN list`
-* existence check: `n.prop IS NOT NULL`
-* range search: `n.prop > value`
-* prefix search: `STARTS WITH`
-
-[NOTE]
-====
-For details about each operator, see xref::syntax/operators.adoc[Operators].
-====
-
-However, predicates might be planned as existence check and a filter.
-For most predicates, this can be avoided by following these restrictions:
-
-* If there is any `equality check` and `list membership check` predicates,
-they need to be for the first properties defined by the index.
-* There can be up to one `range search` or `prefix search` predicate.
-* There can be any number of `existence check` predicates.
-* Any predicate after a `range search`, `prefix search` or `existence check` predicate has to be an `existence check` predicate.
-
-[NOTE]
-====
-The `suffix search` (`ENDS WITH`) and `substring search` (`CONTAINS`) predicates can utilize the index as well.
-However, they are always planned as an existence check and a filter and any predicates following after will therefore also be planned as such.
-====
-
-For example, an index on nodes with `:Label(prop1,prop2,prop3,prop4,prop5,prop6)` and predicates:
-
-[source, cypher, role=test-skip]
-----
-WHERE n.prop1 = 'x' AND n.prop2 = 1 AND n.prop3 > 5 AND n.prop4 < 'e' AND n.prop5 = true AND n.prop6 IS NOT NULL
-----
-
-will be planned as:
-
-[source, cypher, role=test-skip]
-----
-WHERE n.prop1 = 'x' AND n.prop2 = 1 AND n.prop3 > 5 AND n.prop4 IS NOT NULL AND n.prop5 IS NOT NULL AND n.prop6 IS NOT NULL
-----
-
-with filters on `n.prop4 < 'e'` and `n.prop5 = true`, since `n.prop3` has a `range search` predicate.
-
-And an index on nodes with `:Label(prop1,prop2)` with predicates:
-
-[source, cypher, role=test-skip]
-----
-WHERE n.prop1 ENDS WITH 'x' AND n.prop2 = false
-----
-
-will be planned as:
-
-[source, cypher, role=test-skip]
-----
-WHERE n.prop1 IS NOT NULL AND n.prop2 IS NOT NULL
-----
-
-with filters on `n.prop1 ENDS WITH 'x'` and `n.prop2 = false`, since `n.prop1` has a `suffix search` predicate.
-
-Composite indexes require predicates on all properties indexed.
-If there are predicates on only a subset of the indexed properties, it will not be possible to use the composite index.
-To get this kind of fallback behavior, it is necessary to create additional indexes on the relevant sub-set of properties or on single properties.
-
-
 [[indexes-create-indexes]]
 == +CREATE INDEX+
 
@@ -421,36 +353,6 @@ Creating an index requires xref::administration/access-control/database-administ
 ====
 The new index is not immediately available, but is created in the background.
 ====
-
-*Examples:*
-
-* xref::indexes-for-search-performance.adoc#indexes-create-range-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-single-property-range-index-for-nodes[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-single-property-range-index-for-relationships[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-composite-range-index-for-nodes[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-composite-range-index-for-relationships[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-range-index-only-if-it-does-not-already-exist[]
-* xref::indexes-for-search-performance.adoc#indexes-create-token-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-node-label-lookup-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-relationship-type-lookup-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-lookup-index-only-if-it-does-not-already-exist[]
-* xref::indexes-for-search-performance.adoc#indexes-create-point-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-node-point-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-relationship-point-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-point-index-only-if-it-does-not-already-exist[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-point-index-specifying-the-index-configuration[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-point-index-specifying-both-the-index-provider-and-configuration[]
-* xref::indexes-for-search-performance.adoc#indexes-create-text-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-node-text-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-relationship-text-index[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-text-index-only-if-it-does-not-already-exist[]
-** xref::indexes-for-search-performance.adoc#indexes-create-a-text-index-specifying-the-index-provider[]
-* xref::indexes-for-search-performance.adoc#indexes-create-conflicting-index[]
-** xref::indexes-for-search-performance.adoc#indexes-failure-to-create-an-already-existing-index[]
-** xref::indexes-for-search-performance.adoc#indexes-failure-to-create-an-index-with-the-same-name-as-an-already-existing-index[]
-** xref::indexes-for-search-performance.adoc#indexes-failure-to-create-an-index-when-a-constraint-already-exists[]
-** xref::indexes-for-search-performance.adoc#indexes-failure-to-create-an-index-with-the-same-name-as-an-already-existing-constraint[]
-
 
 [[indexes-create-range-index]]
 === Creating a range index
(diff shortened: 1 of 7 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-500.surge.sh

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -572,10 +535,9 @@ CREATE INDEX index_name FOR (n:Label) ON (n.prop1, ..., n.propN)
 ----
```
Hunterness: should we keep the command here or remove it as well?

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -612,10 +573,9 @@ CREATE INDEX index_name FOR ()-[r:TYPE]-() ON (r.prop1, ..., r.propN)
...
```
Hunterness, line 413: should we keep the command here or remove it as well?

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -393,94 +391,96 @@ If there are predicates on only a subset of the indexed properties, it will not
...
+** xref::indexes-for-search-performance.adoc#indexes-create-a-text-index-specifying-the-index-provider[]
+* xref::indexes-for-search-performance.adoc#indexes-create-conflicting-index[]
+** xref::indexes-for-search-performance.adoc#indexes-failure-to-create-an-already-existing-index[]
+** xref::indexes-for-search-performance.adoc#indexes-failure-to-create-an-index-with-the-same-name-as-an-already-existing-index[]
+** xref::indexes-for-search-performance.adoc#indexes-failure-to-create-an-index-when-a-constraint-already-exists[]
+** xref::indexes-for-search-performance.adoc#indexes-failure-to-create-an-index-with-the-same-name-as-an-already-existing-constraint[]
+
+
+[[indexes-create-range-index]]
+=== Creating a range index
```
Hunterness: should these new sections have discrete headers as well?

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -639,79 +599,74 @@ Added 1 index.
...
-====
 
-[discrete]
-[[administration-indexes-create-a-relationship-type-lookup-index]]
-=== Create a relationship type lookup index
+[[indexes-create-token-index]]
+=== Creating a token lookup index
 
-A named relationship type lookup index for all relationships with any relationship type can be created with:
+Creating a token lookup index can be done with the `CREATE LOOKUP INDEX` command.
```
Hunterness:
```suggestion
Creating a token lookup index (node label or relationship type lookup index) can be done with the `CREATE LOOKUP INDEX` command.
```

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -639,79 +599,74 @@ Added 1 index.
...
 
-[source, syntax, role="noheader"]
-----
-CREATE LOOKUP INDEX index_name FOR ()-[r]-() ON EACH type(r)
-----
+Token lookup indexes (node label and relationship type lookup indexes) have only one index provider available, `token-lookup-1.0`, and no supported index configuration.
```
Hunterness:
```suggestion
Token lookup indexes have only one index provider available, `token-lookup-1.0`, and no supported index configuration.
```

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -957,11 +890,11 @@ Added 1 index.
...
 
 Specifying the index configuration can be combined with specifying index provider.
-
+Though only one valid value exists for the index provider, `point-1.0`, which is the default value.
 
 [discrete]
-[[administration-indexes-create-a-point-index-specifying-both-the-index-provider-and-configuration]]
-=== Create a point index specifying both the index provider and configuration
+[[indexes-create-a-point-index-specifying-both-the-index-provider-and-configuration]]
+==== Create a point index specifying both the index provider and configuration
```
Hunterness: Should this example also be removed? since we only have one provider

stefano-ottolenghi: I think so

stefano-ottolenghi (commented):
This looks much, much better! I really like the grouping under _Create index_, and the more considerate use of `[discrete]`.

Some thoughts:
1. I don't support renaming sections, which means breaking anchor links. If there's significant moving around of content that an old anchor doesn't make sense anymore, then fine, but if it's making it more descriptive or dropping `administration` ... I don't think there's enough gain given that the flipside is that links to those break. You have fixed a bunch in this manual, but who knows how many there are scattered around in docs.
1. I'd get rid of the large index that's at the root of `CREATE INDEX`. The main sections/groupings are found in the table of contents in the sidebar, while the individual sections have their own list (which I do like).
1. I would remove the `Syntax` section, break it out, and spread it to the relevant sections (maybe with a `Syntax` subsection when there's a lot, like in `CREATE INDEX`). Especially now with the broken-down show index groupings, we can put the relevant syntax in each group.
  Extra hints that syntax is in the wrong place are that 
  a) all admonitions found there are duplicated in `CREATE INDEX`
  b) `CREATE INDEX` opens by saying `Creating an index is done with the CREATE ... INDEX ... command`, which lacks so much detail to actually be helpful.
  c) it would also give more meaning to sentences like `Range indexes have only one index provider available, range-1.0, and no supported index configuration`, which are now a bit mysterious unless one has read the syntax section upfront.
1. Can we find a way to remove/merge/move at least some of the admonitions? I am looking in particular at `CREATE INDEX`, and having 2 regular (short) paragraphs and 4 admonitions is a sign that we can design it better. Furthermore, almost all examples have the remarks `The index name must be unique.` and many have `Note that the index is not immediately available, but is created in the background.`, which are duplicates of the opening admonitions.
1. Move `Composite index limitations` as last section - I doubt anybody cares about this _before_ reading some of the other sections.
1. I'm torn whether I support the `Result \n Added 1 index.` blocks in examples. I understand that they are there to show that the behavior is different if you append `IF NOT EXISTS`, but since
  a) there's really no variation, other than `Added 1 index` and `(no changes, no records)`
  b) there's a very descriptive admonition in the 20% of examples that exhibit the second result, saying `The index will not be created if there already exists an index with the same schema and type, same name or both.`
... I would take it away. Makes examples shorter.
1. Is the order of sections in `CREATE INDEX` the best? From my limited experience, I thought the best, sorted by popularity, would be `range > text > point > lookup` (since lookup are created by default on install).
2. It's a good time to get rid of all the `Query` titles on examples. Then if we also address 6., we can put the example title as title of the code block, remove the surrounding section `=====` bloat, and shrink the examples size by 50%!

Hunterness:
@stefano-ottolenghi 
1. the main reason that I removed the `administration` is that we now have a chapter called `Administration` which the indexes isn't included in so having administration in the link felt wrong to me. (your argument on link breaking is most likely why it wasn't changed back in 4.3 when the old administration chapter which included indexes was removed.) As for links, I tried searching both in the cypher manual, cheat sheet and operations manual for usages of the links I changed (didn't find any in the operations manual and have a PR up for the cheat sheet ones)
2. sure
3. When we did break the previous administration chapter we also restructured the constraint section into separate pages and it was talk about doing similarly for the index one, that however didn't happen. The constraints section has syntax and examples separated into two pages, which I kinda likes. It feels odd to me to not have all syntax parts collected (we do that for almost all of our teams command documentation, be that database management or indexes), however we quite often repeat the syntax pattern when it's more complicated in the start of the individual sections (for example one big table/section for user commands and then repeat of the create syntax in the create user section). That could maybe be an alternative?
4. `Note that the index is not immediately available, but is created in the background.` I thought I had removed those from each individual example and only had it at the start of the create (and syntax) section? maybe I missed some or remember wrong :shrug: I can take a look at combining them/make them into text
5. I think it has been there since before I started at neo XD but sure I can move it down :P
6. I think that if we want to update something like this it should be consistent with the constraints section, I'd probably do it separately and as part of this (especially since I'm unsure if I agree with removing them)
7. I have no preference to the order, I can switch it around
8. can you give an example for how it should look? I'm unsure I understand (and don't know enough asciidoc XD)

stefano-ottolenghi:
1. I had not checked other manuals, but great! Then totally ahead!
2. .
3. I see the structural _appeal_ that separating syntax from examples may have, but is it what you would _expect_ to find, if you were a person _unfamiliar_ with the topic? Cause I appreciate (and expect) much more a structure like this:
  ```
  Create index
  --/ Create range index
    --/ Syntax (only for range index)
    --/ Examples (only for range index)
--/ Create text index
    --/ Syntax (only for text index)
    --/ Examples (only for text index)
  ```
I am never going to be interested in the syntax of ALL commands _at once_ (unless I am a Cypher or field engineer looking for a quick reference, which we shouldn't take as intended readership).
In general, as a reader, if there is a section called `Create index`, I expect to find all information about index creation _in there_, and not that some bits are somewhere else. Does this resonate with you?

4. Same applies to the index uniqueness remarks :)

6+8. This is how I imagined it:

Screenshot from 2023-04-11 10-19-13

or, if we don't care about the example title since there's already so much info surrounding it,

Screenshot from 2023-04-11 10-20-50

source is
```
==== Create a single-property range index for nodes

The example shows how to create a named range index on all nodes with label `Person` and having the `surname` property:

.Creating a named range index on a single property for all nodes with a particular label
[source, cypher]
----
CREATE INDEX node_range_index_name FOR (n:Person) ON (n.surname)
----

[NOTE]
====
The index name must be unique.
====
```
where `.Creating a named range index on a single property for all nodes with a particular label` was omitted in the second version.

I know the current docs style does not have padding and line underneath h3/h4 headers - I hacked it in the browser, but I'm gonna push it as a permanent css change.

Hunterness:
1. Is there other places I should check as well or is these 3 enough?
2. .
3. On one hand sure, on the other I like how the constraints section is structured and would probably rather do something like that and split it further? I'll make some more re-arranging and we can see if it's better
4. (edit) I checked and I have the comment on availability at the start of CREATE INDEX as well as start of each index section but not in each individual example. Is this good enough or do we want to remove it further?

6-8. I'll see what I can do
(edit): looking over your pictures again, I feel like it makes it seem less like examples? it no longer have the `example XX` text but are just random code? and if I add back syntax summaries in the sections I feel like there isn't really a big difference visually in what is a syntax code-block and what is an example? (more than the special characters in the text)

modules/ROOT/pages/indexes-for-search-performance.adoc
Hunterness: Gonna go with no `discrete` on these

Hunterness:
To keep track of the updates:
- [x] 1
- [x] 2
- [x] 3
- [x] 4
- [x] 5
- [x] 6
- [x] 7
- [x] 8

stefano-ottolenghi:
Great work!!

> Is there other places I should check as well or is these 3 enough?

I think operations manual was the most likely to have. It should be enough.

> On one hand sure, on the other I like how the constraints section is structured and would probably rather do something like that and split it further? I'll make some more re-arranging and we can see if it's better

As I said, we should think more about what inexperienced readers expect, what is best for them, and try to put ourselves in their shoes. The fact that a structure _looks nice_ does not mean that it's pedagogical. ... with that said, I see you have moved syntax inline in each section, so :+1: 

:arrow_right: The only change I'd recommend is breaking in two separate `source` blocks the two syntax options that are now currently in one single block :)

:arrow_right: Also, given the sentence `Range indexes have only one index provider available, range-1.0, and no supported index configuration.`, how about we take away the `configuration` bit from the corresponding syntax blocks, and also this sentence? As it is, it feels like we are showing something to users in the syntax, and then telling them: "btw, you can't do anything with that". The fact that the backend code _accepts_ that parameter doesn't mean that we should document it, if it's not of interest to users.

> I checked and I have the comment on availability at the start of CREATE INDEX as well as start of each index section but not in each individual example. Is this good enough or do we want to remove it further?

I find repetition noisy unless extremely necessary. If you deem if necessary then go for it, but I think we can assume that people will read the `CREATE INDEX` intro, so they can get that piece of info if they want. There's also not so many use cases where the <5 seconds before an index is ready are a game-changer, so I think we can bless readers with a sentence less for each section :D

> 6-8. I'll see what I can do

:+1: 

> looking over your pictures again, I feel like it makes it seem less like examples? it no longer have the example XX text but are just random code? and if I add back syntax summaries in the sections I feel like there isn't really a big difference visually in what is a syntax code-block and what is an example? (more than the special characters in the text)

There's not too much visual difference, true. To preserve it we can either 

- use `Syntax` as title of the syntax blocks, so that they get the blue bulky header, AND not give a title to examples, so they DON'T get the bulky header
- viceversa: no title to syntax blocks (as now), but title to all examples

:arrow_right: We could also make a `Syntax` section everywhere where relevant, `discrete` of course.

:arrow_right: One more think I can spot is that `Examples` is always just a paragraph in bold. I'd recommend it making a discrete subsection as well.

Hunterness:
> Also, given the sentence Range indexes have only one index provider available, range-1.0, and no supported index configuration., how about we take away the configuration bit from the corresponding syntax blocks, and also this sentence?

I think you mean the `options` part, and I can remove that on the indexes where it's not relevant. However I think the sentence should stay since the provider and (empty) config map gets returned in the show indexes and should therefore at least be mentioned in the docs somewhere.

> with that said, I see you have moved syntax inline in each section, so 👍

Yes, each section gets a syntax summary of relevant syntax parts, but I'm also keeping the syntax section for easy comparison and will leave the full syntax there.

stefano-ottolenghi:
This is looking really good! My last thought is whether we should move the full syntax section to the bottom, either before or after the composite index remarks? If it is for reference/quick glance, as you said, I think it makes more sense that it comes after the more content-heavy sections that people will be interested into first.

modules/ROOT/pages/indexes-for-search-performance.adoc
Hunterness: @stefano-ottolenghi do you have any opinions around these?

stefano-ottolenghi: Ah, good point. I think we can get rid of those. There's the full syntax in two places, and plenty of examples for people to infer what the syntax for each particular case looks like.

stefano-ottolenghi:
I was reviewing the style changes I've done locally and I spotted a couple glitches:

- `Create a text index only if it does not already exist` is missing one equal sign in front
- `Only one node label lookup index can exist at the time.` admonitions: I think they'd be better placed _after_ each example, rather than before. Also, what does it mean, exactly? Is it a typo and should be `at a time` (as in _simultaneously)?
- Also, in case you missed it, https://github.com/neo4j/docs-cypher/pull/500#discussion_r1165121439

stefano-ottolenghi (approved):
Awesome! 
Is there something else you'd like to change, or shall I merge?

Hunterness:
I think it's good to merge from my view👍

But didn't you have issues in the docs team with tests/TC not being run?

stefano-ottolenghi:
> But didn't you have issues in the docs team with tests/TC not being run?

Yours is one of the mysterious blessed PRs that got picked up correctly in TC (you can see the successful status of `pull-requests-cypher-test-dev-neo4j-dev` here).

PR #506, opened 2023-04-13
[Testing] Correct query plan for some operators (rollback)

stefano-ottolenghi:
To a good extent, this is a rollback of #395.

This investigation started with the query plan of `CacheProperties` being wrong since... well, there was no `CacheProperties` operator in the plan. The testing was happy that way, which was odd. It turned out that it was because of the example query in the `Eager` operator, which effectively _cleared_ the database, so that all examples afterwards ran on an empty db. `CacheProperties` comes after `Eager` in the operators page, so that the `CacheProperties` example ran on an empty db.

Summary of changes:
- Changed `Eager` example so that it deletes only _two_ nodes, rather than _all_ nodes. The query plan is essentially unchanged, other than indexes being used. Most notably, `Eager` is still triggered, presumably as the db doesn't know upfront how many nodes have that property value anyway.
- Fixing the above triggered a bunch of failures (for `CacheProperties`, `EagerAggregation`, `Distinct`) , which are essentially the plans changed in #395. I've restored them as they were before, but not altered them otherwise. 
- Given proper names to indexes/constraints, describing what they _index_ rather than what they _test_.
- Moved the setup queries _inside_ the sections they belong to, rather than _before_ the respective openings.
- Removed a setup block that dropped and recreated the exact same index.
- The `Delete` example was acting on `(me:Person {name: 'me'})`, which has relationships attached to it, so an error is raised. Or rather, no failure used to happen because, previously, the `Eager` example cleared the whole db, so `DELETE me` was deleting a non-existent node, and now the `Eager` example `DETACH DELETE`s the `me` node, so that it keeps not existing. However, if for whatever reason we changed some of the examples involved and it turned out that `me` existed with relationships (as the page setup creates), then we'd get a failure. So I've changed the `DELETE` example to act on `you`, rather than `me`, as `you` never exists. There's also some perverse joy involved with pretending to obliterate the reader, so I like it better.


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-506.surge.sh

modules/ROOT/pages/execution-plans/operators.adoc
```diff
@@ -1155,17 +1151,18 @@ Total database accesses: 2, total allocated memory: 184
...
+
+[[query-plan-node-unique-index-seek]]
+== Node Unique Index Seek
+// NodeUniqueIndexSeek
+
 ////
 [source, cypher, role=test-setup]
 ----
-CREATE CONSTRAINT constraint_Team_name IF NOT EXISTS FOR (t:Team) REQUIRE (t.name) IS UNIQUE
+CREATE CONSTRAINT Team_name IF NOT EXISTS FOR (t:Team) REQUIRE (t.name) IS UNIQUE
```
JPryce-Aklundh: Should 'Team' be capitalised? Same question for below example.

modules/ROOT/pages/execution-plans/operators.adoc
```diff
@@ -3724,19 +3719,21 @@ Runtime version {neo4j-version-minor}
...
++------------------+----+----------------------------+----------------+------+---------+----------------+------------------------+-----------+------------+---------------------+
+| Operator         | Id | Details                    | Estimated Rows | Rows | DB Hits | Memory (Bytes) | Page Cache Hits/Misses | Time (ms) | Ordered by | Pipeline            |
++------------------+----+----------------------------+----------------+------+---------+----------------+------------------------+-----------+------------+---------------------+
+| +ProduceResults  |  0 | l                          |             14 |    6 |      12 |                |                    1/0 |     0.049 |            |                     |
+| |                +----+----------------------------+----------------+------+---------+----------------+------------------------+-----------+            |                     |
+| +OrderedDistinct |  1 | l                          |             14 |    6 |       0 |             32 |                    0/0 |     0.049 |            | In Pipeline 0       |
```
JPryce-Aklundh: This is not the correct operator for the example.

JPryce-Aklundh (commented):
This looks great! The example for the distinct operator does not include the correct operator, though.

modules/ROOT/pages/execution-plans/operators.adoc
stefano-ottolenghi: No strong opinion on this, but makes sense to have it lowercase.

modules/ROOT/pages/execution-plans/operators.adoc
stefano-ottolenghi: Ouch, good spot!

I changed it from `RETURN l` to `RETURN p` and that also fixes the operators list. The reason is that:

> The planner picks a NodeByLabelScan on l as the starting point for this query. NodeByLabelScan gives rows in order of the node ids.
If the expression after a DISTINCT happens to be ordered, the planner can choose an OrderedDistinct operator instead of a Distinct operator. The 2 operators are doing the same thing, but OrderedDistinct will consume way less memory.
When you have DISTINCT p, on the other hand, the rows are still ordered by l and not by p. So there the planner has to choose a Distinct operator.

JPryce-Aklundh (approved):

PR #527, opened 2023-04-28
[TESTING] Ensure string type results are surrounded by quotes

stefano-ottolenghi:
Testing infrastructure now checks that strings are ... really strings! Meaning they need to be surrounded by some form of quotes (either single or double). Couple of other mistakes fixed along the way.


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-527.surge.sh

PR #539, opened 2023-05-04
Add documentation license and remove reference to beginners in introduction 

JPryce-Aklundh:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-539.surge.sh

davidoliverSP2 (commented):
Thanks @JPryce-Aklundh 

I think this small chapter could benefit from a few other changes:

1. I'm not sure there is benefit from having this section titled - I don't think people would search for "intended readership", and it looks a little awkward that we may be filtering people _out_ of the docs before they start. I would suggest removing the header "Cypher Manual: intended readership" and letting the remaining paragraphs flow from the previous content.
2. This sentence looks out of place: `it assumes a degree of prior knowledge of Cypher` - this statement is true, but perhaps could be removed, and incorporated into the final paragraph.
3. As it stands, the tone of the final paragraph seems a little out of place - I would suggest a small tweak, (which incorporates point 2 above):
> If you are new to Cypher and Neo4j, you can visit the [Getting Started Guide -> Introduction to Cypher](https://neo4j.com/docs/getting-started/cypher-intro/) chapter for concepts and deeper background. Additionally, [Neo4j GraphAcademy](https://graphacademy.neo4j.com/) has a variety of free courses tailored for all levels of experience.

davidoliverSP2 (approved):
thanks @JPryce-Aklundh 👍

PR #543, opened 2023-05-09
Add graph.propertiesByName query example that selects subset of graphs

smietana:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-543.surge.sh

modules/ROOT/pages/functions/graph.adoc
```diff
@@ -89,6 +89,22 @@ Properties for all graphs on the current composite database are returned.
...
+WHERE any(tag IN props.tags WHERE tag IN ["A"])
+CALL {
+  USE graph.byName(name)
+  MATCH (n)
+  RETURN n
+}
+RETURN n
+----
+
+Returns all nodes from a subset of graphs that have a tags property containing "A".
```
renetapopova:
```suggestion
Returns all nodes from a subset of graphs that have a `tags` property containing `"A"`.
```

renetapopova (approved):
Thanks, @smietana. A small formatting suggestion from my side.

modules/ROOT/pages/functions/graph.adoc
renetapopova: @smietana, into which version should this PR go?

smietana: Function graph.propertiesByName was added in 5.2, so if we have minor version granularity it should go to 5.2 and above otherwise 5.x.

modules/ROOT/pages/functions/graph.adoc
```diff
@@ -89,6 +89,22 @@ Properties for all graphs on the current composite database are returned.
 
 |===
 
+.Query
+[source, cypher, indent=0]
+----
+UNWIND graph.names() AS name
+WITH name, graph.propertiesByName(name) AS props
+WHERE any(tag IN props.tags WHERE tag IN ["A"])
```
tobias-johansson:
```suggestion
WHERE "A" IN props.tags
```
Since `tags` is a list we could just check if it contains `"A"` like this right?

smietana: Right, this is equivalent

tobias-johansson (approved):
:+1:

PR #548, opened 2023-05-10
Remove reference to `dbms.cluster.overview`

jackwaudby:
Users should now use `SHOW SERVERS` as of 5.x.

### DIFF
--- a/modules/ROOT/pages/administration/databases.adoc
+++ b/modules/ROOT/pages/administration/databases.adoc
@@ -176,7 +176,7 @@ The default for a standalone database is `neo4j://localhost:7687`. label:default
 | STRING
 
 | writer
-|`true` for the database node that accepts writes (this node is the leader for this database in a cluster or this is a standalone instance). label:default-output[]
+|`true` for the instance that accepts writes for this database (this instance is the leader for this database in a cluster or this is a standalone instance). label:default-output[]
 | BOOLEAN
 
 | requestedStatus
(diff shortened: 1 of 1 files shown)


jackwaudby:
This change should apply to all 5.x versions.

lidiazuin (approved):

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -707,10 +707,10 @@ Controls how the system handles existing data on disk when creating the database
 Currently this is only supported with `existingDataSeedInstance` and must be set to `use` which indicates the existing data files should be used for the new database.
 
 | `existingDataSeedInstance`
-| instance ID of the cluster node
+| server ID of the cluster node
```
AlexicaWright: We're trying to move away from using _node_ in this context. We don't use node in the operations manual so we should be consistent with that. I'd say something like: _ID of the cluster server_ or _server ID of the cluster instance_.
There are two more _node_s in this chapter that also need changing. I'll add a suggestion.

AlexicaWright: I can't leave an inline comment, but this is on line 179:
`true` for the database node that accepts writes (this node is the leader for this database in a cluster or this is a standalone instance).
I suggest changing this to:
`true` for the database instance that accepts writes (this instance is the leader for this database in a cluster or this is a standalone instance).

AlexicaWright (commented):
Just an editorial comment ;)

modules/ROOT/pages/administration/databases.adoc
jackwaudby: Sure thing! I've made the changes and pushed up. Feel free to hit merge

neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-548.surge.sh

AlexicaWright (approved):
👍 LGTM thanks for updating!

PR #549, opened 2023-05-10
Mark Cypher first usage on every page

recrwplay:
Add the `mark-terms` extension to mark the first instance of Cypher on every page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@neo4j-antora/mark-terms": "1.0.0",
     "@neo4j-documentation/macros": "^1.0.2",
     "@neo4j-documentation/remote-include": "^1.0.0",
-    "asciidoctor-kroki": "^0.17.0"
+    "asciidoctor-kroki": "^0.16.0"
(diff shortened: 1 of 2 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-549.surge.sh

davidoliverSP2:
Nice

Is there a way for it to ignore headers? it looks a little odd in the navigation

Screenshot 2023-05-12 at 11 50 58

recrwplay:
> Is there a way for it to ignore headers? it looks a little odd in the navigation

Yes, or with a ui-bundle update we should be able to hide them from view when titles are used in navigation.

recrwplay:
I've updated the extension so that in version 1.1.0 you can specify whether to mark or ignore titles (and it also recognises list items). This PR now builds without marking usage in titles.

davidoliverSP2 (approved):
Great stuff, thanks @recrwplay 👍

PR #554, opened 2023-05-12
Fix misaligned entries in the TOC

davidoliverSP2:
It's likely that we won't need an appendix soon, especially given that right now there is only 1 entry in it, but this PR fixes the misalignment in the TOC for that chapter.


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-554.surge.sh

recrwplay (approved):
LGTM

PR #561, opened 2023-05-16
Add NODES 23 ad

recrwplay:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-561.surge.sh

davidoliverSP2:
@recrwplay the surge link (+ local builds) doesn't pick up the ad - should it? (could this PR depend on another in in the UI repo?)

recrwplay:
I added it to the `publish.yml` playbook only, and the checks here use `preview.yml`. You could test it locally by building with the publish playbook, but you'd also need to update the source branches in that playbook, replacing `5.x` with `HEAD`

davidoliverSP2 (approved):
Looks good, thanks @recrwplay 👍 

Lets add a reminder to remove this on the 27th Oct!

PR #583, opened 2023-05-29
Cleanup redundant asciidoc formatting from clauses examples

stefano-ottolenghi:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-583.surge.sh

modules/ROOT/pages/clauses/index.adoc
```diff
@@ -169,7 +169,7 @@ These comprise sub-clauses that must operate as part of reading clauses.
 m| xref::clauses/where.adoc[WHERE]
 | Adds constraints to the patterns in a `MATCH` or `OPTIONAL MATCH` clause or filters the results of a `WITH` clause.
 
-m| xref::clauses/order-by.adoc[+ORDER BY [ASC[ENDING] \| DESC[ENDING]]+]
+m| xref::clauses/order-by.adoc[ORDER BY [ASC[ENDING] \| DESC[ENDING]]]
```
JPryce-Aklundh: This does not render properly 

Screenshot 2023-05-29 at 11 53 10

JPryce-Aklundh (approved):
Thanks a lot @stefano-ottolenghi ! I just noticed one thing that did not render as it should

modules/ROOT/pages/clauses/index.adoc
stefano-ottolenghi: Ah, good catch! The brackets needed escaping (already before, but the issue was hidden by the extra `+`).

PR #588, opened 2023-06-01
Documentation updates for clarifying what is allowed in LIMIT & Co

arnefischereit:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-588.surge.sh

modules/ROOT/pages/clauses/call-subquery.adoc
```diff
@@ -534,7 +538,9 @@ The amount of work to do in each separate transaction can be specified in terms
 to process before committing the current transaction and starting a new one.
 The number of input rows is set with the modifier `OF n ROWS` (or `ROW`).
 If omitted, the default batch size is `1000` rows.
-The following is the same example but with one transaction every `2` input rows:
+The number of rows can be expressed using any expression that evaluates to a positive integer -- however the expression cannot refer to nodes or relationships.
+
+The following is the same example as above for loading a CSV file but with one transaction every `2` input rows:
```
AzuObs: I am not sure whether we should backquote `2` (and also`1000`) a couple lines above. But I leave it up to you or the documentation team to decide on style.

AzuObs (approved):
Nice job. The comment is optional 😄

modules/ROOT/pages/clauses/call-subquery.adoc
```diff
@@ -534,7 +538,9 @@ The amount of work to do in each separate transaction can be specified in terms
 to process before committing the current transaction and starting a new one.
 The number of input rows is set with the modifier `OF n ROWS` (or `ROW`).
 If omitted, the default batch size is `1000` rows.
-The following is the same example but with one transaction every `2` input rows:
+The number of rows can be expressed using any expression that evaluates to a positive integer -- however the expression cannot refer to nodes or relationships.
```
JPryce-Aklundh:
```suggestion
The number of rows can be expressed using any expression that evaluates to a positive integer and does not refer to nodes or relationships.
```

modules/ROOT/pages/clauses/call-subquery.adoc
JPryce-Aklundh:
```suggestion
This example loads a CSV file with one transaction for every `2` input rows:
```

JPryce-Aklundh: Both should be within backticks (`1000` because it is referring to a default value, and `2` because it references something in a code example).

modules/ROOT/pages/clauses/limit.adoc
```diff
@@ -59,7 +59,7 @@ d|Rows: 3
 [[limit-subset-rows-using-expression]]
 == Using an expression with `LIMIT` to return a subset of the rows
 
-Limit accepts any expression that evaluates to a positive integer as long as it is not referring to any external variables:
+`LIMIT` accepts any expression that evaluates to a positive integer, as long as it can be statically calculated:
```
JPryce-Aklundh: "statically calculated" should probably be briefly defined - what do you think about either of the below 2 suggestions?

```suggestion
`LIMIT` accepts any expression that evaluates to a positive integer, as long as it can be calculated before the query is run."
```

Alternatively: "...as long as it can be statically calculated (i.e. calculated before the query is run)."

modules/ROOT/pages/clauses/skip.adoc
```diff
@@ -87,7 +87,7 @@ d|Rows: 2
 [[skip-using-expression]]
 == Using an expression with `SKIP` to return a subset of the rows
 
-Skip accepts any expression that evaluates to a positive integer as long as it is not referring to any external variables:
+`SKIP` accepts any expression that evaluates to a positive integer, as long as it can be statically calculated:
```
JPryce-Aklundh: See above comment.

modules/ROOT/pages/clauses/call-subquery.adoc
arnefischereit: So, you do not deem the information important, that this is only a slight modification of the previous example?

JPryce-Aklundh: Not really, no. I think it works as a stand-alone example, without tempting the reader to scroll back up again.

JPryce-Aklundh (approved):

Issue #542, opened 2023-05-06
"shadowning of variable not allowed", a doc bug?


complyue:
I would perceive from context of this doc, that "variable shadowing" may happen, but could be undesirable when not intended. Is the "not allowed" phrasing a doc bug here?

https://github.com/neo4j/docs-cypher/blob/727287b89881ba1ab1697a619c079527209e912a/modules/ROOT/pages/syntax/expressions.adoc?plain=1#LL553C1-L553C61

stefano-ottolenghi:
As the docs point out, shadowing is _not_ allowed in those scenarios. I have expanded the two relevant examples to clarify a bit more and to show what error you would get, and how you would address it :)

PR #591, opened 2023-06-05
Improve `WITH` shadowing examples to clarify behavior

stefano-ottolenghi:
I have slightly changed the wording as a result of #542, and also expanded the first two relevant examples so that the error you would get is shown (as was already the case in the section `COLLECT subquery with WITH`).


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-591.surge.sh

PR #596, opened 2023-06-09
Add general information around the new types and their synonyms

Hunterness:
This will be referenced in the property type constraint and type predicate expression documentation.

### DIFF
--- a/modules/ROOT/pages/syntax/operators.adoc
+++ b/modules/ROOT/pages/syntax/operators.adoc
@@ -409,7 +409,7 @@ For example, `1 > b` and `1 < b` are both false when b is NaN.
   *** xref::syntax/expressions.adoc#cypher-expressions-general[`String`]
   *** xref::syntax/expressions.adoc#cypher-expressions-general[`Boolean`]
   *** xref::syntax/expressions.adoc#cypher-expressions-general[`Number`]
- ** The value `null` is considered larger than any value.
+ ** The value `null` is ordered after all other values.
 * *Ordering* of constructed type values:
  ** For the xref::values-and-types/property-structural-constructed.adoc#constructed-types[constructed types] (e.g. maps and lists), elements of the containers are compared pairwise for ordering and thus determine the ordering of two container types.
 For example, `[1, 'foo', 3]` is ordered before `[1, 2, 'bar']` since `'foo'` is ordered before `2`.
--- a/modules/ROOT/pages/syntax/operators.adoc
+++ b/modules/ROOT/pages/syntax/operators.adoc
@@ -395,20 +395,20 @@ For example, `1 > b` and `1 < b` are both false when b is NaN.
 * Comparing for ordering when one argument is `null` (e.g. `null < 3` is `null`).
 * *Ordering* of values with *different* types:
  ** The ordering is, in ascending order, defined according to the following list:
-  *** xref::values-and-types/maps.adoc#cypher-literal-maps[`Map`]
-  *** xref::values-and-types/property-structural-constructed.adoc#structural-types[`Node`]
-  *** xref::values-and-types/property-structural-constructed.adoc#structural-types[`Relationship`]
-  *** xref::values-and-types/lists.adoc[`List`]
-  *** xref::syntax/patterns.adoc#cypher-pattern-path-variables[`Path`]
-  *** xref::values-and-types/temporal.adoc[`DateTime`]
-  *** xref::values-and-types/temporal.adoc[`LocalDateTime`]
-  *** xref::values-and-types/temporal.adoc[`Date`]
-  *** xref::values-and-types/temporal.adoc[`Time`]
-  *** xref::values-and-types/temporal.adoc[`LocalTime`]
-  *** xref::values-and-types/temporal.adoc[`Duration`]
-  *** xref::syntax/expressions.adoc#cypher-expressions-general[`String`]
-  *** xref::syntax/expressions.adoc#cypher-expressions-general[`Boolean`]
-  *** xref::syntax/expressions.adoc#cypher-expressions-general[`Number`]
+  *** xref::values-and-types/maps.adoc#cypher-literal-maps[`MAP`]
+  *** xref::values-and-types/property-structural-constructed.adoc#structural-types[`NODE`]
+  *** xref::values-and-types/property-structural-constructed.adoc#structural-types[`RELATIONSHIP`]
+  *** xref::values-and-types/lists.adoc[`LIST`]
+  *** xref::syntax/patterns.adoc#cypher-pattern-path-variables[`PATH`]
+  *** xref::values-and-types/temporal.adoc[`ZONED DATETIME`]
+  *** xref::values-and-types/temporal.adoc[`LOCAL DATETIME`]
+  *** xref::values-and-types/temporal.adoc[`DATE`]
+  *** xref::values-and-types/temporal.adoc[`ZONED TIME`]
+  *** xref::values-and-types/temporal.adoc[`LOCAL TIME`]
+  *** xref::values-and-types/temporal.adoc[`DURATION`]
+  *** xref::syntax/expressions.adoc#cypher-expressions-general[`STRING`]
+  *** xref::syntax/expressions.adoc#cypher-expressions-general[`BOOLEAN`]
+  *** Numbers: xref::syntax/expressions.adoc#cypher-expressions-general[`INTEGER`, `FLOAT`]
  ** The value `null` is ordered after all other values.
 * *Ordering* of constructed type values:
  ** For the xref::values-and-types/property-structural-constructed.adoc#constructed-types[constructed types] (e.g. maps and lists), elements of the containers are compared pairwise for ordering and thus determine the ordering of two container types.
--- a/modules/ROOT/pages/syntax/operators.adoc
+++ b/modules/ROOT/pages/syntax/operators.adoc
@@ -331,7 +331,7 @@ xref::clauses/where.adoc#query-where-string[String matching] contains more infor
 
 ==== Equality
 
-Cypher supports comparing values (see xref::values-and-types/property-structural-composite.adoc[]) by equality using the `=` and `<>` operators.
+Cypher supports comparing values (see xref::values-and-types/property-structural-constructed.adoc[]) by equality using the `=` and `<>` operators.
 
 Values of the same type are only equal if they are the same identical value (e.g. `3 = 3` and `"x" <> "xy"`).
 
@@ -396,8 +396,8 @@ For example, `1 > b` and `1 < b` are both false when b is NaN.
 * *Ordering* of values with *different* types:
  ** The ordering is, in ascending order, defined according to the following list:
   *** xref::values-and-types/maps.adoc#cypher-literal-maps[`Map`]
-  *** xref::values-and-types/property-structural-composite.adoc#structural-types[`Node`]
-  *** xref::values-and-types/property-structural-composite.adoc#structural-types[`Relationship`]
+  *** xref::values-and-types/property-structural-constructed.adoc#structural-types[`Node`]
+  *** xref::values-and-types/property-structural-constructed.adoc#structural-types[`Relationship`]
   *** xref::values-and-types/lists.adoc[`List`]
   *** xref::syntax/patterns.adoc#cypher-pattern-path-variables[`Path`]
   *** xref::values-and-types/temporal.adoc[`DateTime`]
@@ -410,8 +410,8 @@ For example, `1 > b` and `1 < b` are both false when b is NaN.
   *** xref::syntax/expressions.adoc#cypher-expressions-general[`Boolean`]
   *** xref::syntax/expressions.adoc#cypher-expressions-general[`Number`]
  ** The value `null` is considered larger than any value.
-* *Ordering* of composite type values:
- ** For the xref::values-and-types/property-structural-composite.adoc#composite-types[composite types] (e.g. maps and lists), elements of the containers are compared pairwise for ordering and thus determine the ordering of two container types.
+* *Ordering* of constructed type values:
+ ** For the xref::values-and-types/property-structural-constructed.adoc#constructed-types[constructed types] (e.g. maps and lists), elements of the containers are compared pairwise for ordering and thus determine the ordering of two container types.
 For example, `[1, 'foo', 3]` is ordered before `[1, 2, 'bar']` since `'foo'` is ordered before `2`.
 
 
@@ -609,17 +609,17 @@ The following table shows -- for each combination of operation and operand type 
 
 | xref::syntax/operators.adoc#syntax-multiply-divide-duration-number[`*`]
 | _Duration_
-| xref::values-and-types/property-structural-composite.adoc#property-types[Number]
+| xref::values-and-types/property-structural-constructed.adoc#property-types[Number]
 | _Duration_
 
 | xref::syntax/operators.adoc#syntax-multiply-divide-duration-number[`*`]
-| xref::values-and-types/property-structural-composite.adoc#property-types[Number]
+| xref::values-and-types/property-structural-constructed.adoc#property-types[Number]
 | _Duration_
 | _Duration_
 
 | xref::syntax/operators.adoc#syntax-multiply-divide-duration-number[`/`]
 | _Duration_
-| xref::values-and-types/property-structural-composite.adoc#property-types[Number]
+| xref::values-and-types/property-structural-constructed.adoc#property-types[Number]
 | _Duration_
 
 |===
--- a/modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
+++ b/modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
@@ -75,7 +75,7 @@ These types (and their synonyms) can be used in type predicate expressions and p
 However, not all types can be used in all places.
 // TODO: add link to type predicate expression section (once exists)
 // TODO: add link to property type constraint section (once exists)
-Some of these types are not yet supported in either type predicate expressions nor property type constraints.
+Some of these types are not yet supported in either type predicate expressions or property type constraints.
 // TODO: Remove sentence about type not being supported once all types are allowed
 
 [.synonyms, opts="header", cols="2a,2a"]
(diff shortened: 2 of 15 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-596.surge.sh

modules/ROOT/pages/syntax/operators.adoc
```diff
@@ -395,23 +395,23 @@ For example, `1 > b` and `1 < b` are both false when b is NaN.
...
+  *** xref::values-and-types/temporal.adoc[`DATE`]
+  *** xref::values-and-types/temporal.adoc[`ZONED TIME`]
+  *** xref::values-and-types/temporal.adoc[`LOCAL TIME`]
+  *** xref::values-and-types/temporal.adoc[`DURATION`]
+  *** xref::syntax/expressions.adoc#cypher-expressions-general[`STRING`]
+  *** xref::syntax/expressions.adoc#cypher-expressions-general[`BOOLEAN`]
+  *** Numbers: xref::syntax/expressions.adoc#cypher-expressions-general[`INTEGER`, `FLOAT`]
+ ** The value `null` is ordered after all other values.
+* *Ordering* of constructed type values:
+ ** For the xref::values-and-types/property-structural-constructed.adoc#constructed-types[constructed types] (e.g. maps and lists), elements of the containers are compared pairwise for ordering and thus determine the ordering of two container types.
```
l-heemann, line 414: Should we mention what happens if one constructed type has fewer elements than the other (and other than that is a subtype?)

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
```diff
@@ -50,23 +50,55 @@ In Neo4j, all relationships have a direction.
 However, you can have the notion of undirected relationships at query time.
 ====
 
-[[composite-types]]
-== Composite types
+[[constructed-types]]
+== Constructed types
```
l-heemann, line 54: How long have we referred to this as "composite types?" Should we have a "formerly known as composite types" or do we have the ability to insert SEO tags here?

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
```diff
@@ -50,23 +50,55 @@ In Neo4j, all relationships have a direction.
...
 
+[[types-synonyms]]
+== Types and their synonyms
+
+The table below shows the types and their syntactic synonyms.
+These types (and the synonyms) can be used in the type predicate expressions and the property type constraints.
+However, not all types can be used in all places.
+// TODO: add link to type predicate expression section (once exists)
+// TODO: add link to property type constraint section (once exists)
+Some of these types are not yet supported in either type predicate expressions nor property type constraints.
```
l-heemann: I believe "nor" is a gramatical error here?
"A or B" -> `A || B`
"Either A or B" -> `A ^ B` (xor)
"Neither A nor B" -> `!A && !B`

I believe what you are trying to write is that they are all supported in at least one of them, but only a subset are supported in both? In that case I would write 
```suggestion
Some of these types are not yet supported in either type predicate expressions or property type constraints.
```

l-heemann (approved):

modules/ROOT/pages/syntax/operators.adoc
Hunterness: Not sure, this is an older comment (we just updated the name from composite to constructed types), I think generally if you compare something pairwise and they don't have the same amount of values then the shorter will come first :shrug:

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
Hunterness: No at the moment what we want to say is that a subset is at the moment not covered in any of them (there are for example nodes and maps which did not make the cut for the expressions either)

So it's probably more of
```suggestion
Some of these types are not yet supported in neither type predicate expressions nor property type constraints.
```

modules/ROOT/pages/syntax/operators.adoc
l-heemann: There is an argument to be made that this isn't obvious together with "The value `null` is ordered after all other values". I.e. a list of size 2 may be seen as (a, b, null, null, null) but I guess that is far fetched.

gem-neo4j: But a list in Cypher can contain nulls, so a list of [a, b, null, null, null] would be considered size 5 right?

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
```diff
@@ -1,47 +1,47 @@
...
-These fall into the following three categories: *property*, *structural*, and *composite*. 
+These fall into the following three categories: *property*, *structural*, and *constructed*.
 This section will first provide a brief overview of each type, and then go into more detail about the property data type. 
 
 == Property types
 
 A property type value is one that can be stored as a node or relationship property.
 
-The following data types are included in the property types category: `Boolean`, `Date`, `DateTime`, `Duration`, `Float`, `Integer`, `List`, `LocalTime` `LocalDateTime`, `Point`, `String`, and `Time`.
+The following data types are included in the property types category: `BOOLEAN`, `DATE`, `DURATION`, `FLOAT`, `INTEGER`, `LIST`, `LOCAL DATETIME`, `LOCAL TIME`, `POINT`, `STRING`, `ZONED DATETIME`, and `ZONED TIME`.
```
gem-neo4j, line 15: Shall we order this according to CIP-100?

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
```diff
@@ -50,23 +50,55 @@ In Neo4j, all relationships have a direction.
...
 
-Composite values can also contain `null`. 
+Constructed type values can also contain `null`.
 For more details, see xref::values-and-types/working-with-null.adoc[working with null].
 
+[[types-synonyms]]
+== Types and their synonyms
+
+The table below shows the types and their syntactic synonyms.
+These types (and the synonyms) can be used in the type predicate expressions and the property type constraints.
```
gem-neo4j:
```suggestion
These types (and their synonyms) can be used in type predicate expressions and property type constraints.
```

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
gem-neo4j: That sounds wrong to me ^^

I don't think you should be using neither/nor when you have not already in the sentence. So Lasse's suggestion of either/or sounds better to me (just a feeling, don't know the grammar rules)

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
```diff
@@ -50,23 +50,55 @@ In Neo4j, all relationships have a direction.
...
+However, not all types can be used in all places.
+// TODO: add link to type predicate expression section (once exists)
+// TODO: add link to property type constraint section (once exists)
+Some of these types are not yet supported in either type predicate expressions nor property type constraints.
+// TODO: Remove sentence about type not being supported once all types are allowed
+
+[.synonyms, opts="header", cols="2a,2a"]
+|===
+| Type | Synonyms
+| `BOOLEAN` | `BOOL`
```
gem-neo4j, line 84: Can we order this to match the ordering in CIP 100?

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
```diff
@@ -50,23 +50,55 @@ In Neo4j, all relationships have a direction.
...
+| `DATE` |
+| `DURATION` |
+| `FLOAT` |
+| `INTEGER` | `INT`, `SIGNED INTEGER`
+| `LIST<INNER_TYPE>` | `ARRAY<INNER_TYPE>`, `INNER_TYPE LIST`, `INNER_TYPE ARRAY`
```
gem-neo4j, line 89: We might also allow `ANY<INTEGER | FLOAT> LIST` but not sure if I'd call that an inner type?

modules/ROOT/pages/values-and-types/spatial.adoc
```diff
@@ -17,18 +17,17 @@ For more information about the comparison and ordering of spatial values, see th
...
+Values with the `POINT` type have the following characteristics:
 
 * Each point can have either 2 or 3 dimensions. 
 This means it contains either 2 or 3 64-bit floating point values, which together are called the _Coordinate_.
-* Each point will also be associated with a specific xref::values-and-types/spatial.adoc#cypher-spatial-crs[Coordinate Reference System] (CRS) that determines the meaning of the values in the _Coordinate_.
-* Instances of `Point` and lists of `Point` can be assigned to node and relationship properties.
-* Nodes and relationships with `Point` or `List(Point)` properties can be indexed using a xref::indexes-for-search-performance.adoc#indexes-create-a-node-point-index
-[point index]. 
+* Each point will also be associated with a specific xref::values-and-types/spatial.adoc#spatial-values-crs[Coordinate Reference System] (CRS) that determines the meaning of the values in the _Coordinate_.
+* Instances of `POINT` and `LIST<POINT>` can be assigned to node and relationship properties.
```
gem-neo4j, line 29: I guess this is pending the decision on nullable inner types for constraints? as LIST<POINT> isn't technically assignable to a node, it should be LIST<POINT NOT NULL>

modules/ROOT/pages/values-and-types/temporal.adoc
```diff
@@ -21,21 +21,35 @@ The following table lists the temporal value types and their supported component
...
-By contrast, `Duration` is not a temporal instant type.
-A `Duration` represents a temporal amount, capturing the difference in time between two instants, and can be negative.
-`Duration` captures the amount of time between two instants, it does not capture a start time and end time.
+By contrast, `DURATION` is not a temporal instant type.
+A `DURATION` represents a temporal amount, capturing the difference in time between two instants, and can be negative.
+`DURATION` captures the amount of time between two instants, it does not capture a start time and end time.
 
+Some temporal types have been renamed slightly.
+The table below shows the current as well as the old names of the temporal types.
+_This switch was introduced in Neo4j 5.9._
```
gem-neo4j:
```suggestion
_This change was introduced in Neo4j 5.9._
```

gem-neo4j (commented):
Looks good, just a few comments :D Cool to clean up the docs!

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
Hunterness: let's see what the docs team thinks then before updating

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
Hunterness: it is currently alphabetical order (which is the same as it was before), I don't really have a preference in either direction

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
Hunterness: A lot of this page uses alphabetical ordering so we kept that for now. Once unions are allowed we planned to add something about the order and potentially some general normalization if there is something we needed to document. If you feel strongly we could reorder the list but a lot of places in this chapter uses alphabetical and it's kinda helpful when looking for something specific :shrug:

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
Hunterness: We currently don't have anything around unions as they are not part of the already listed types in this chapter nor used in the constraints or expressions. Once they are the union type will be just another type that can be the inner type of a list, and probably also mention that if you use those variants of the list type you need the ANY<...> grouping.

modules/ROOT/pages/values-and-types/spatial.adoc
Hunterness: in this case I guess `LIST<POINT NOT NULL>`, but I don't think we need to distinguish it here? like sure you cant store lists with nulls but this is mostly just pointing out that you can have lists of points or points in your properties (we just updated the type names). and until we introduce NOT NULL for types I don't think we should add it here.

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
Lojjs: I was very unsure about the grammatics here when we wrote it 🙈 But the intention is what Therese said

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
gem-neo4j: eh alphabetical corder is fine probably :)

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
gem-neo4j: I'm not too fussed, it can stay as it is :)

modules/ROOT/pages/values-and-types/spatial.adoc
gem-neo4j: Sure thing :)

gem-neo4j (approved):
I think I only have English points left, which docs can help with anyway :)

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
Hunterness: Not sure if the docs team have things they'd need/want/use to do for this kind of renames, maybe the docs reviewer can answer this?

I could always add a small note if that is the best solution :)

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
stefano-ottolenghi: tl;dr don't mix negative sentences and `neither`, which is another negation.
See here https://english.stackexchange.com/questions/138624/use-of-either-or-in-a-negative-phrase
So Lasse's suggestion is correct.

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
stefano-ottolenghi: Changing the title _and_ the section ID (i.e. the `[[ ... ]]`) will affect the internal search, meaning that people looking for `composite types` will not find it anymore, but I think they will survive. I guess people mostly look for the specific types (maps/lists) rather than the group name.

modules/ROOT/pages/values-and-types/temporal.adoc
```diff
@@ -21,21 +21,35 @@ The following table lists the temporal value types and their supported component
...
+| `DATE`          | {check-mark} |              |
+| `LOCAL TIME`     |              | {check-mark} |
+| `ZONED TIME`          |              | {check-mark} | {check-mark}
+| `LOCAL DATETIME` | {check-mark} | {check-mark} |
+| `ZONED DATETIME`      | {check-mark} | {check-mark} | {check-mark}
+| `DURATION`      | `-`          | `-`          | `-`
 |===
 
-`Date`, `Time`, `LocalTime`, `DateTime`, and `LocalDateTime` are _temporal instant_ types.
+`DATE`, `LOCAL TIME`, `ZONED TIME`, `LOCAL DATETIME`, and `ZONED DATETIME` are _temporal instant_ types.
```
stefano-ottolenghi, line 32: Does all this renaming affect users in any way? 

1. Do users ever directly interact with type names in Cypher?
2. I suppose drivers will need to be updated to work with the new types. Have they already been updated? What happens if a user is running, say, driver version 5.3 with the latest neo4j with renamed types?

modules/ROOT/pages/values-and-types/temporal.adoc
```diff
@@ -21,21 +21,35 @@ The following table lists the temporal value types and their supported component
...
-By contrast, `Duration` is not a temporal instant type.
-A `Duration` represents a temporal amount, capturing the difference in time between two instants, and can be negative.
-`Duration` captures the amount of time between two instants, it does not capture a start time and end time.
+By contrast, `DURATION` is not a temporal instant type.
+A `DURATION` represents a temporal amount, capturing the difference in time between two instants, and can be negative.
+`DURATION` captures the amount of time between two instants, it does not capture a start time and end time.
 
+Some temporal types have been renamed slightly.
+The table below shows the current as well as the old names of the temporal types.
+_This change was introduced in Neo4j 5.9._
```
stefano-ottolenghi: `slightly` is irrelevant - the amount of implementation work to comply with new names is the same regardless of the size of the renaming. I'd remove that.

I'd actually group the first and last sentence as "Starting from Neo4j 5.9, some temporal types have been renamed."

modules/ROOT/pages/values-and-types/temporal.adoc
```diff
@@ -717,7 +731,7 @@ RETURN d.timezone, d.offset, d.offsetMinutes, d.epochSeconds, d.epochMillis
 .+date.truncate+
 ======
 
-Get the first _Date_ of the current year:
+Get the first date of the current year:
```
stefano-ottolenghi: I suppose this should say "first _day_", as also the return name? First _dates_ are usually much more romantic than our docs.

modules/ROOT/pages/values-and-types/temporal.adoc
Lojjs: 1. Only in the new features property type constraints and type predicate expressions which are shipped in 5.9 together with this documentation, not before. 
2. There are no new types, only introducing names in Cypher of existing ones, so I don't think this will impact drivers

Hunterness: +1 on Louise answer

We haven't previously had named types as such, which is why I too believe drivers will be unaffected by this

modules/ROOT/pages/values-and-types/temporal.adoc
Hunterness:
```suggestion
Starting from Neo4j 5.9, some temporal types have been renamed.
The table below shows the current as well as the old names of the temporal types.
```

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
Hunterness: Then we'll leave as is

modules/ROOT/pages/values-and-types/temporal.adoc
stefano-ottolenghi:
```suggestion
Get the first day of the current year:
```

stefano-ottolenghi (approved):
:bird:

PR #601, opened 2023-06-10
Update unwind.adoc

sh-ghanique:
Better illustration that `UNWIND [] AS empty` stops executing the query...

Because `empty` is also part of the RETURN statement, it could be falsely assumed that that is the reason why the query returns an empty list.


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-601.surge.sh

JPryce-Aklundh (approved):
Thanks @sh-ghanique - we appreciate your input!

PR #607, opened 2023-06-13
Deprecate self-references in CREATE clauses

hedengran:


modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -16,6 +16,31 @@ New features are added to the language continuously, and occasionally, some feat
 This section lists all of the features that have been removed, deprecated, added, or extended in different Cypher versions.
 Replacement syntax for deprecated and removed features are also indicated.
 
+[[cypher-deprecations-additions-removals-5.8]]
```
sherfert:
```suggestion
[[cypher-deprecations-additions-removals-5.9]]
```

neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-607.surge.sh

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -16,6 +16,31 @@ New features are added to the language continuously, and occasionally, some feat
...
+label:functionality[]
+label:deprecated[]
+[source, cypher, role="noheader"]
+----
+CREATE (a {foo:1}), (b {foo:a.foo})
+----
+a|
+
+Creating a node or relationship entity,
+and then referencing a property of that entity in the same create clause is deprecated.
```
sherfert:
```suggestion
and then referencing that entity in the same create clause is deprecated.
```

This is not really about the property part, right? Any other reference, e.g. `labels(a)` would also be deprecated?

sherfert (commented):
Sorry, I had this pending for a while but forgot to submit the review.

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
hedengran: You're right, "referencing a property" does not catch it all, forgot about those other cases. But some references are still valid, e.g. `CREATE (a)-[:REL]->(a)`. Not sure how to capture that in one sentence.

sherfert: Does "referencing that entity in a property definition in the same create clause ..." catch all the cases?

hedengran: Yeah I think that works, thanks!

sherfert (approved):

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -19,6 +19,29 @@ Replacement syntax for deprecated and removed features are also indicated.
...
+label:deprecated[]
+[source, cypher, role="noheader"]
+----
+CREATE (a {foo:1}), (b {foo:a.foo})
+----
+a|
+
+Creating a node or relationship entity,
+and then referencing that entity in a property definition in the same create clause is deprecated.
+Please split your `CREATE` into two separate clauses instead.
```
```suggestion
and then referencing that entity in a property definition in the same `CREATE` clause is deprecated.
Split the `CREATE` clause into two separate clauses instead.
```

JPryce-Aklundh (changes requested):
looks good overall 👍 - I made two suggestions:

JPryce-Aklundh (approved):

PR #597, opened 2023-06-09
Add property type constraints

Hunterness:
Adds links to places added in https://github.com/neo4j/docs-cypher/pull/596 (which will be broken until that goes in).

- [x] Fix one of the TODOs introduced in the PR above (linking to this section in one of the new places there)

PR #614, opened 2023-06-14
Document type predicate expression

Lojjs:
Builds on top of https://github.com/neo4j/docs-cypher/pull/597


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-614.surge.sh

Lojjs:
[type predicate expression](https://trello.com/c/GSnWmtev/608-type-predicate-expression)

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -14,7 +14,7 @@ This section contains an overview of expressions in Cypher with examples.
 [NOTE]
 ====
 Most expressions in Cypher evaluate to `null` if any of their inner expressions are `null`.
-Notable exceptions are the operators `IS NULL` and `IS NOT NULL`.
+Notable exceptions are the operators `IS NULL` and `IS NOT NULL` and the type predicate expression.
```
Hunterness:
```suggestion
Notable exceptions are the operators `IS NULL`, `IS NOT NULL`, and the type predicate expression.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,187 @@ RETURN COLLECT {
...
+* `ZONED DATETIME`
+* `DURATION`
+* `POINT`
+
+Allowed syntax variations of these types are listed xref::values-and-types/property-structural-constructed.adoc#types-synonyms[here].
+
+.Query
+[source, cypher]
+----
+UNWIND [42, true, 'abc'] AS val
```
Hunterness, line 1266: should you also show null? Might be good to mention that it will be true for all types (until we have NOT NULL at least :P)

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,187 @@ RETURN COLLECT {
...
+| +42+ | +true+
+| +true+ | +false+
+| +'abc'+ | +false+
+2+d|Rows: 3
+|===
+
+[[type-predicate-expressions-not]]
+=== Type predicate expressions with NOT
+
+It is also possible to verify that a Cypher expression does not have a certain type, using `IS NOT ::`.
```
Hunterness: it feels odd to just have `IS NOT ::`, feels like the type is missing?

modules/ROOT/pages/syntax/expressions.adoc
Hunterness: found it below

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,187 @@ RETURN COLLECT {
...
+
+[[type-predicate-expressions-null]]
+=== Type predicate expressions for null
+All Cypher types are considered to include the `null` value.
+Therefore, `IS ::` will return `true` for all expressions evaluating to `null`, regardless of the type.
+
+.Query
+[source, cypher]
+----
+RETURN toFloat(null) IS :: BOOLEAN AS isBoolean
```
Hunterness: I guess the precedence of these 2 examples is that the toFloat and addition is evaluated first, but do we want to clarify that with parentesis?

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,187 @@ RETURN COLLECT {
...
+| +name+ | +age+
+| +Charlie+ | +21+
+2+d|Rows: 1
+|===
+
+
+[[type-predicate-expressions-number-sizes]]
+=== Type predicate expressions for numbers of different sizes
+For numerical values passed in as parameters, Cypher does not take the size of the number into account.
+Cypher will thus regard any integral parameter as an `INTEGER`, regardless of if it is declared as e.g. an INT16 or an INT64 in the client programming language.
```
Hunterness: is `integral` the correct word here?

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -47,6 +47,29 @@ It returns a string representation of the property type for property type constr
...
+a|
+label:functionality[]
+label:new[]
+[source, cypher, role="noheader"]
+----
+IS [NOT] :: <TYPE>
+----
+a|
+
+Added type predicate expressions. The available types are:
```
gem-neo4j: Should this be linked to the docs about them for easy access?

modules/ROOT/pages/syntax/expressions.adoc
Lojjs: Maybe... Currently we have no documentation about precedence whatsoever, I have a card on our wall for that

modules/ROOT/pages/keyword-glossary.adoc
```diff
@@ -1034,6 +1034,9 @@ xref::functions/temporal/index.adoc#functions-temporal-truncate-overview[Truncat
 
 | xref:syntax/expressions.adoc#existential-subqueries[EXISTS {...}]
 | Evaluates the existence of a subquery.
+
+| xref:syntax/expressions.adoc#type-predicate-expressions[IS :: `type`]
+| Verifies that an expression has a certain type.
```
gem-neo4j: Not sure if has is correct, would "Verifies that an expression is of a certain type" be better? Or is an instance of a certain type

modules/ROOT/pages/syntax/expressions.adoc
gem-neo4j:
```suggestion
It is also possible to verify that a Cypher expression does not have a certain type, using the negated type expression; `IS NOT ::`.
```

perhaps ^^

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,187 @@ RETURN COLLECT {
...
+| +true+
+1+d|Rows: 1
+|===
+
+Likewise, `IS NOT ::` will return `false` for all expressions evaluating to `null`, regardless of the type.
+
+.Query
+[source, cypher]
+----
+RETURN null + 1 IS NOT :: DATE AS isNotDate
```
gem-neo4j: Maybe parenthesis would help here as well

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,187 @@ RETURN COLLECT {
...
+| +isNotDate+
+| +false+
+1+d|Rows: 1
+|===
+
+[[type-predicate-expressions-properties]]
+=== Type predicate expression for properties
+Type predicate expressions can also be used to filter out nodes or relationships with properties of a certain type.
+
+Assume that we have a graph containing the following nodes:
```
gem-neo4j: Do we normally use personal pronouns in the docs?

modules/ROOT/pages/syntax/expressions.adoc
gem-neo4j: "any exact numerical parameter" ?

gem-neo4j: Also they need to fit into an INT64 still right?

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,187 @@ RETURN COLLECT {
...
+| +isInteger+
+| +true+
+1+d|Rows: 1
+|===
+
+For more information on parameters, see xref::syntax/parameters.adoc[].
+
+[[type-predicate-expressions-variations]]
+=== Syntactical variations of type predicate expressions
+Type predicate expressions allows for some alternative syntax:
```
gem-neo4j:
``` suggestion
Type predicate expressions allow for some alternative syntax:
```

gem-neo4j (commented):
Looks good! just a few comments :)

modules/ROOT/pages/syntax/expressions.adoc
Lojjs: Good catch, I will check how it is written above

modules/ROOT/pages/syntax/expressions.adoc
Lojjs: Let me add parentheses without mentioning precedence for now

Hunterness: feels more needed for the addition part so you don't think the type expression applies to the 1 and then gets added to null, the function call at least keeps itself together better

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
Hunterness: should also be added to the one for constraints then (and I know this is one of the lines I've updated in my PR though not to add a link)

modules/ROOT/pages/syntax/expressions.adoc
Hunterness: I made one comment for both but agree, it probably helps more here than fr the function call

modules/ROOT/pages/syntax/expressions.adoc
Lojjs: I guess that is technically true. Not sure how to formulate that the number should be of a valid range to not overflow

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -98,6 +98,29 @@ a|
...
+a|
+label:functionality[]
+label:new[]
+[source, cypher, role="noheader"]
+----
+IS [NOT] :: <TYPE>
+----
+a|
+
+Added xref:syntax/expressions.adoc#type-predicate-expressions[type predicate expressions]. The available types are:
```
Hunterness: two things
1. could you do the same for the constraint one below? it didn't get to be a part of the constraint PR but would be nice
2. since I got this on my PR :P
```suggestion
Added xref:syntax/expressions.adoc#type-predicate-expressions[type predicate expressions].
The available types are:
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,187 @@ RETURN COLLECT {
...
+| +name+ | +age+
+| +Charlie+ | +21+
+2+d|Rows: 1
+|===
+
+
+[[type-predicate-expressions-number-sizes]]
+=== Type predicate expressions for numbers of different sizes
+For numerical values passed in as parameters, Cypher does not take the size of the number into account.
+Cypher will thus regard any exact numerical parameter as an `INTEGER`, regardless of if it is declared as e.g. an INT16 or an INT64 in the client programming language.
```
Hunterness: This part is kinda hard to read: `regardless of if it is declared as e.g.`. It feels like it starts out one way and then halfway though switches to another?

gem-neo4j: Cypher will therefore regard any exact numerical parameter as an `INTEGER` regardless of its declared size e.g. an INT16 or an INT32 passed through from a client programming language will both be treated as an `INTEGER`.

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,186 @@ RETURN COLLECT {
...
+_This feature was introduced in Neo4j 5.9._
+
+A type predicate expression can be used to verify the type of a variable, literal, property or other Cypher expression.
+
+[source, syntax, role="noheader", indent=0]
+----
+<expr> IS :: <TYPE>
+----
+
+where `<expr>` is any Cypher expression and `<TYPE>` is one of the following types:
```
JPryce-Aklundh:
```suggestion
Where `<expr>` is any Cypher expression and `<TYPE>` is one of the following types:
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,186 @@ RETURN COLLECT {
...
+| +42+ | +true+
+| +true+ | +false+
+| +'abc'+ | +false+
+2+d|Rows: 3
+|===
+
+[[type-predicate-expressions-not]]
+=== Type predicate expressions with NOT
+
+It is also possible to verify that a Cypher expression is not of a certain type, using the negated type predicate expression; `IS NOT ::`.
```
JPryce-Aklundh: Does the double colons have to be included?

```suggestion 
It is also possible to verify that a Cypher expression is not of a certain type, using the negated type predicate expression `IS NOT ::`.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,186 @@ RETURN COLLECT {
...
+| +val+ | +notString+
+| +42+ | +true+
+| +true+ | +true+
+| +'abc'+ | +false+
+2+d|Rows: 3
+|===
+
+[[type-predicate-expressions-null]]
+=== Type predicate expressions for null
+All Cypher types are considered to include the `null` value.
```
JPryce-Aklundh: My suggestion may be overly simplistic, but "considered to include" is a bit cryptic I think. 

Also, please include a space after a header (applies to below headers as well).  
```suggestion

All Cypher types include the `null` value.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,186 @@ RETURN COLLECT {
...
+[source, cypher, role=test-setup]
+----
+MATCH (n) DETACH DELETE n;
+CREATE (:Person {name: 'Alice', age:14}),
+(:Person {name:'Bob', age:'20'}),
+(:Person {name:'Charlie', age:21}),
+----
+////
+
+image:graph_expression_type_predicate.svg[]
```
JPryce-Aklundh, line 1355: I will update this with an image created in Arrows in a later PR.

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,186 @@ RETURN COLLECT {
...
+MATCH (n:Person)
+WHERE n.age IS :: INTEGER AND n.age > 18
+RETURN n.name AS name, n.age AS age
+----
+
+.Result
+[role="queryresult",options="header,footer",cols="2*<m"]
+|===
+| +name+ | +age+
+| +Charlie+ | +21+
```
JPryce-Aklundh: `+` not required when `m` is used in column defintion. 
```suggestion
| name | age
| Charlie | 21
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,186 @@ RETURN COLLECT {
...
+
+image:graph_expression_type_predicate.svg[]
+
+.Query
+[source, cypher]
+----
+MATCH (n:Person)
+WHERE n.age IS :: INTEGER AND n.age > 18
+RETURN n.name AS name, n.age AS age
+----
```
JPryce-Aklundh, line 1365:
```suggestion
The following query finds all `Person` nodes with an `age` property that is an `INTEGER` with a greater value than `18`.

.Query
[source, cypher]
----
MATCH (n:Person)
WHERE n.age IS :: INTEGER AND n.age > 18
RETURN n.name AS name, n.age AS age
----
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,186 @@ RETURN COLLECT {
...
+| +name+ | +age+
+| +Charlie+ | +21+
+2+d|Rows: 1
+|===
+
+
+[[type-predicate-expressions-number-sizes]]
+=== Type predicate expressions for numbers of different sizes
+For numerical values passed in as parameters, Cypher does not take the size of the number into account.
+Cypher will therefore regard any exact numerical parameter as an `INTEGER` regardless of its declared size e.g. an `INT16` or an `INT32` passed through from a client programming language will both be treated as an `INTEGER`.
```
JPryce-Aklundh: Final sentence added as a suggestion following previous review-comments.

```suggestion
Cypher will therefore regard any exact numerical parameter as an `INTEGER` regardless of its declared size.
For example, an `INT16` or an `INT32` passed through from a client programming language will both be treated by Cypher as an `INTEGER`.
Note that any exact numerical parameter used must fit within the range of an `INT 64`. 
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1232,6 +1232,186 @@ RETURN COLLECT {
...
+
+.Result
+[role="queryresult",options="header,footer",cols="1*<m"]
+|===
+| +isInteger+
+| +true+
+1+d|Rows: 1
+|===
+
+For more information on parameters, see xref::syntax/parameters.adoc[].
```
JPryce-Aklundh:
```suggestion
More information about parameters can be found xref::syntax/parameters.adoc[here].
```

modules/ROOT/pages/values-and-types/working-with-null.adoc
```diff
@@ -4,7 +4,7 @@
 = Working with `null`
 
 In Cypher, `null` is used to represent missing or undefined values.
-All data types in Cypher are nullable. 
+All data types in Cypher are nullable, which leads to the xref::syntax/expressions.adoc#type-predicate-expressions-null[type predicate expression] always  returning `true` for `null` values.
```
JPryce-Aklundh:
```suggestion
All data types in Cypher are nullable.
This means that xref::syntax/expressions.adoc#type-predicate-expressions-null[type predicate expressions] always return `true` for `null` values.
```

JPryce-Aklundh (commented):
Thanks @Lojjs - I have a few comments/suggestions, but looks very good overall.
(It will be even better once we dig it out of the current syntax/expressions page and give it a more prominent place in a new expressions chapter 😄 )

modules/ROOT/pages/syntax/expressions.adoc
Lojjs: Yes, the versions of the syntax is 
x IS :: STRING (preferred)
x IS TYPED STRING
x :: STRING

x IS NOT :: STRING (preferred)
x IS NOT TYPED STRING

modules/ROOT/pages/syntax/expressions.adoc
Lojjs: Thanks, I just copied one of the svg images from earlier on the page and modified it

JPryce-Aklundh (approved):

PR #631, opened 2023-06-28
Add example of list syntax to server management commands

nick-giles-neo:
Had a user try to use the SET OPTIONS { tags:foo,bar,baz }, so being a little more helpful in description.

PR #632, opened 2023-06-28
Add example of list syntax to server management commands

renetapopova:
Had a user try to use the SET OPTIONS { tags:foo,bar,baz }, so being a little more helpful in description.
Cherry-picked from #631


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-632.surge.sh

PR #634, opened 2023-06-28
Type predicate expressions iteration 2

gem-neo4j:
Type predicate expressions now contain non-property value types as well as NOT NULL variants and LISTs.

### DIFF
--- a/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
+++ b/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
@@ -16,6 +16,51 @@ New features are added to the language continuously, and occasionally, some feat
 This section lists all of the features that have been removed, deprecated, added, or extended in different Cypher versions.
 Replacement syntax for deprecated and removed features are also indicated.
 
+[[cypher-deprecations-additions-removals-5.10]]
+== Version 5.10
+
+===  Updated features
+
+a|
+label:functionality[]
+label:updated[]
+[source, cypher, role="noheader"]
+----
+IS [NOT] :: <TYPE>
+----
+a|
+
+Added xref:syntax/expressions.adoc#type-predicate-expressions[type predicate expressions].
+The newly supported types are:
+
+* `NOTHING`
+* `NULL`
+* `BOOLEAN NOT NULL`
+* `STRING NOT NULL`
+* `INTEGER NOT NULL`
+* `FLOAT NOT NULL`
+* `DATE NOT NULL`
+* `LOCAL TIME NOT NULL`
+* `ZONED TIME NOT NULL`
+* `LOCAL DATETIME NOT NULL`
+* `ZONED DATETIME NOT NULL`
+* `DURATION NOT NULL`
+* `POINT NOT NULL`
+* `NODE`
+* `NODE NOT NULL`
+* `RELATIONSHIP`
+* `RELATIONSHIP NOT NULL`
+* `MAP`
+* `MAP NOT NULL`
+* `LIST<TYPE>`
+* `LIST<TYPE> NOT NULL`
+* `PATH`
+* `PATH NOT NULL`
+* `PROPERTY VALUE`
+* `PROPERTY VALUE NOT NULL`
+* `ANY`
+* `ANY NOT NULL`
+
 [[cypher-deprecations-additions-removals-5.9]]
 == Version 5.9
 
--- a/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
+++ b/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
@@ -30,7 +30,7 @@ IS [NOT] :: <TYPE>
 ----
 a|
 
-Added xref:syntax/expressions.adoc#type-predicate-expressions[type predicate expressions].
+Extended xref:syntax/expressions.adoc#type-predicate-expressions[type predicate expressions].
 The newly supported types are:
 
 * `NOTHING`
--- a/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
+++ b/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
@@ -21,6 +21,11 @@ Replacement syntax for deprecated and removed features are also indicated.
 
 ===  Updated features
 
+[cols="2", options="header"]
+|===
+| Feature
+| Details
+
 a|
 label:functionality[]
 label:updated[]
@@ -60,6 +65,8 @@ The newly supported types are:
 * `PROPERTY VALUE NOT NULL`
 * `ANY`
 * `ANY NOT NULL`
+
+|===
 
 [[cypher-deprecations-additions-removals-5.9]]
 == Version 5.9
--- a/modules/ROOT/pages/syntax/expressions.adoc
+++ b/modules/ROOT/pages/syntax/expressions.adoc
@@ -147,7 +147,6 @@ END
 |===
 
 
-.Query
 [source, cypher]
 ----
 MATCH (n)
@@ -159,7 +158,6 @@ CASE n.eyes
 END AS result
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +result+
@@ -204,7 +202,6 @@ END
 | If no match is found, `default` is returned.
 |===
 
-.Query
 [source, cypher]
 ----
 MATCH (n)
@@ -216,7 +213,6 @@ CASE
 END AS result
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +result+
@@ -235,7 +231,6 @@ END AS result
 Owing to the close similarity between the syntax of the two forms, sometimes it may not be clear at the outset as to which form to use.
 We illustrate this scenario by means of the following query, in which there is an expectation that `age_10_years_ago` is `-1` if `n.age` is `null`:
 
-.Query
 [source, cypher]
 ----
 MATCH (n)
@@ -251,7 +246,6 @@ This is because a comparison is made between `n.age` and `n.age IS NULL`.
 As `n.age IS NULL` is a boolean value, and `n.age` is an integer value, the `WHEN n.age IS NULL THEN -1` branch is never taken.
 This results in the `ELSE n.age - 10` branch being taken instead, returning `null`.
 
-.Result
 [role="queryresult",options="header,footer",cols="2*<m"]
 |===
 | +n.name+ | +age_10_years_ago+
@@ -265,7 +259,6 @@ This results in the `ELSE n.age - 10` branch being taken instead, returning `nul
 
 The corrected query, behaving as expected, is given by the following generic `CASE` form:
 
-.Query
 [source, cypher]
 ----
 MATCH (n)
@@ -278,7 +271,6 @@ END AS age_10_years_ago
 
 We now see that the `age_10_years_ago` correctly returns `-1` for the node named `Daniel`.
 
-.Result
 [role="queryresult",options="header,footer",cols="2*<m"]
 |===
 | +n.name+ | +age_10_years_ago+
@@ -298,7 +290,6 @@ You can use the result of `CASE` to set properties on a node or relationship.
 For example, instead of specifying the node directly, you can set a property for a node selected by an expression:
 
 
-.Query
 [source, cypher]
 ----
 MATCH (n)
@@ -313,7 +304,6 @@ SET n.colourCode = colourCode
 
 For more information about using the `SET` clause, see xref::clauses/set.adoc[SET].
 
-.Result
 [role="queryresult",options="footer",cols="1*<m"]
 |===
 1+|(empty result)
@@ -333,7 +323,6 @@ When using the simple `CASE` form, it is useful to remember that in Cypher `null
 
 For example, you might expect `age_10_years_ago` to be `-1` for the node named `Daniel`:
 
-.Query
 [source, cypher]
 ----
 MATCH (n)
@@ -346,7 +335,6 @@ END AS age_10_years_ago
 
 However, as `null = null` does not yield `true`, the `WHEN null THEN -1` branch is never taken, resulting in the `ELSE n.age - 10` branch being taken instead, returning `null`.
 
-.Result
 [role="queryresult",options="header,footer",cols="2*<m"]
 |===
 | +n.name+ | +age_10_years_ago+
@@ -412,7 +400,6 @@ In this regard, `EXISTS` subqueries are different from `CALL` subqueries, xref::
 The following example shows this:
 
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -422,7 +409,6 @@ WHERE EXISTS {
 RETURN person.name AS name
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +name+
@@ -437,8 +423,6 @@ RETURN person.name AS name
 A `WHERE` clause can be used in conjunction to the `MATCH`.
 Variables introduced by the `MATCH` clause and the outside scope can be used in this scope.
 
-
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -449,7 +433,6 @@ WHERE EXISTS {
 RETURN person.name AS name
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +name+
@@ -466,7 +449,6 @@ The nesting also affects the scopes.
 That means that it is possible to access all variables from inside the subquery which are either from the outside scope or defined in the very same subquery.
 
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -480,7 +462,6 @@ WHERE EXISTS {
 RETURN person.name AS name
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +name+
@@ -495,7 +476,6 @@ RETURN person.name AS name
 Here the result is a boolean that shows whether the subquery can find the given pattern.
 
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -504,7 +484,6 @@ RETURN person.name AS name, EXISTS {
 } AS hasDog
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="2*<m"]
 |===
 | +name+ | +hasDog+
@@ -524,7 +503,6 @@ _This feature was introduced in Neo4j 5.3._
 It is worth noting that if one branch has a `RETURN` clause, then all branches require one.
 The below example demonstrates that if one of the `UNION` branches was to return at least one row, the entire `EXISTS` expression will evaluate to true.
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -537,7 +515,6 @@ RETURN
     } AS hasPet
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="2*<m"]
 |===
 | +name+        | +hasPet+
@@ -558,7 +535,6 @@ To avoid confusion, shadowing of these variables is not allowed.
 An outside scope variable is shadowed when a newly introduced variable within the inner scope is defined with the same variable.
 In the example below, the outer variable `name` is shadowed and will therefore throw an error.
 
-.Query
 [source, cypher, role=test-fail]
 ----
 WITH 'Peter' as name
@@ -581,7 +557,6 @@ New variables can be introduced into the subquery, as long as they use a differe
 In the example below, a `WITH` clause introduces a new variable.
 Note that the outer scope variable `person` referenced in the main query is still available after the `WITH` clause.
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -593,7 +568,6 @@ WHERE EXISTS {
 RETURN person.name AS name
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +name+
@@ -611,7 +585,6 @@ _This feature was introduced in Neo4j 5.3._
 need to be aliased, which is different compared to xref::clauses/call-subquery.adoc[`CALL` subqueries].
 Any variables returned in an `EXISTS` subquery will not be available after the subquery.
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -622,7 +595,6 @@ WHERE EXISTS {
 RETURN person.name AS name
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +name+
@@ -651,7 +623,6 @@ In this regard, `COUNT` subqueries are different from `CALL` subqueries, xref::c
 The following query exemplifies this and outputs the owners of more than one dog:
 
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -659,7 +630,6 @@ WHERE COUNT { (person)-[:HAS_DOG]->(:Dog) } > 1
 RETURN person.name AS name
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +name+
@@ -675,7 +645,6 @@ A `WHERE` clause can be used inside the `COUNT` pattern.
 Variables introduced by the `MATCH` clause and the outside scope can be used in this scope.
 
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -686,7 +655,6 @@ WHERE COUNT {
 RETURN person.name AS name
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +name+
@@ -703,7 +671,6 @@ _This feature was introduced in Neo4j 5.3._
 `UNION ALL` clauses do not require the `RETURN` clause. However, it is worth noting that if one branch has a `RETURN` clause, then all require one.
 The below example shows the count of pets each person has by using a `UNION` clause:
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -718,7 +685,6 @@ RETURN
     } AS numPets
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="2*<m"]
 |===
 | +name+        | +numPets+
@@ -739,7 +705,6 @@ To avoid confusion, shadowing of these variables is not allowed.
 An outside scope variable is shadowed when a newly introduced variable within the inner scope is defined with the same variable.
 In the example below, the outer variable `name` is shadowed and will therefore throw an error.
 
-.Query
 [source, cypher, role=test-fail]
 ----
 WITH 'Peter' as name
@@ -762,7 +727,6 @@ New variables can be introduced into the subquery, as long as they use a differe
 In the example below, a `WITH` clause introduces a new variable.
 Note that the outer scope variable `person` referenced in the main query is still available after the `WITH` clause.
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -774,7 +738,6 @@ WHERE COUNT {
 RETURN person.name AS name
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +name+
@@ -793,7 +756,6 @@ See a few examples below:
 ===== Using `COUNT` in `RETURN`
 
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person)
@@ -801,7 +763,6 @@ RETURN person.name, COUNT { (person)-[:HAS_DOG]->(:Dog) } as howManyDogs
 
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="2*<m"]
 |===
 | +person.name+ | +howManyDogs+
@@ -816,7 +777,6 @@ RETURN person.name, COUNT { (person)-[:HAS_DOG]->(:Dog) } as howManyDogs
 ===== Using `COUNT` in `SET`
 
 
-.Query
 [source, cypher]
 ----
 MATCH (person:Person) WHERE person.name ="Andy"
@@ -825,7 +785,6 @@ RETURN person.howManyDogs as howManyDogs
 
 ----
 
-.Result
 [role="queryresult",options="header,footer",cols="1*<m"]
 |===
 | +howManyDogs+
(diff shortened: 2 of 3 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-634.surge.sh

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1257,6 +1259,14 @@ Where `<expr>` is any Cypher expression and `<TYPE>` is one of the following typ
 * `ZONED DATETIME`
 * `DURATION`
 * `POINT`
+* `NODE`
+* `RELATIONSHIP`
+* `MAP`
+* `LIST<TYPE>`
```
gem-neo4j, line 1265: I got stuck with this one, adoc didn't let me do LIST<<TYPE>>, ideas on how to show this?

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -16,6 +16,51 @@ New features are added to the language continuously, and occasionally, some feat
...
+a|
+label:functionality[]
+label:updated[]
+[source, cypher, role="noheader"]
+----
+IS [NOT] :: <TYPE>
+----
+a|
+
+Added xref:syntax/expressions.adoc#type-predicate-expressions[type predicate expressions].
```
Hunterness: you didn't really add the expression though, you just added more allowed types to it?

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1246,6 +1246,8 @@ A type predicate expression can be used to verify the type of a variable, litera
 
 Where `<expr>` is any Cypher expression and `<TYPE>` is one of the following types:
```
Hunterness: If these are all the existing types, could this list not just go away/be combined with the sentence below? I think it's point was to tell which of the types given in `xref::values-and-types/property-structural-constructed.adoc#types-synonyms[types and their synonyms]` could be used but if all of them are now allowed that isn't really that different than just saying something along the lines of 'all the types an their synonyms are supported'

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1257,6 +1259,14 @@ Where `<expr>` is any Cypher expression and `<TYPE>` is one of the following typ
...
+* `NODE`
+* `RELATIONSHIP`
+* `MAP`
+* `LIST<INNER_TYPE>`
+* `PATH`
+* `PROPERTY VALUE`
+* `ANY`
+* `<TYPE> NOT NULL`
 
 For all available Cypher types, see the section on xref::values-and-types/property-structural-constructed.adoc#types-synonyms[types and their synonyms].
```
Hunterness, line 1178: We definitely need to update `values-and-types/property-structural-constructed.adoc` to say things about `NOT NULL` as it is currently missing but should be covered as part of the type system.

There is also a sentence in there with a TODO that can be acted on now that the expressions allow all types listed there.

Hunterness: I can definitely join you on this part if you want

modules/ROOT/pages/syntax/expressions.adoc
Hunterness: Or am I remembering things wrong @Lojjs?

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1302,36 +1312,55 @@ RETURN val, val IS NOT :: STRING AS notString
 [[type-predicate-expressions-null]]
 === Type predicate expressions for null
 
-All Cypher types include the `null` value.
-Therefore, `IS ::` will return `true` for all expressions evaluating to `null`, regardless of the type.
+A Cypher type will include the `null` value unless it is prepended explicitly with `NOT NULL`.
```
Hunterness: not `prepended` I think, I believe that would be something like `NOT NULL INTEGER` instead of `INTEGER NOT NULL`. I think maybe `appended` (`suffixed`?) or something is a better choice

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1302,36 +1312,55 @@ RETURN val, val IS NOT :: STRING AS notString
...
 |===
-| isBoolean
-| true
-1+d|Rows: 1
+| isBoolean | isNotNullBoolean
+| true      | false
+2+d|Rows: 1
+|===
+
+Likewise, `IS NOT ::` will return `false` for all expressions evaluating to `null`, if the type is not prepended with `NOT NULL`.
```
Hunterness: same here, `prepended`

modules/ROOT/pages/syntax/expressions.adoc
Hunterness: speaking of this file, we need to update that list of types and synonyms with the newly introduced types as well (it has some of them such as map and node but not all, for example missing nothing and any. it's just covering the types already mentioned in the types chapter and not the things that are now official types that didn't exist previously.

gem-neo4j: I updated it a little, but we can go over it together tomorrow if you are free :)

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
```diff
@@ -88,15 +87,21 @@ Some of these types are not yet supported in either type predicate expressions o
...
-| `RELATIONSHIP` | `EDGE`
+| `PROPERTY VALUE` | `ANY PROPERTY VALUE`
+| `RELATIONSHIP` | `ANY RELATIONSHIP`, `EDGE`, `ANY EDGE`
 | `STRING` | `VARCHAR`
 | `ZONED DATETIME` | `TIMESTAMP WITH TIMEZONE`
 | `ZONED TIME` | `TIME WITH TIMEZONE`
 |===
 
+All Cypher Types contain the `null` value, in order to make them not nullable `NOT NULL` can be appended to the end of the type.
+e.g `BOOLEAN NOT NULL`, `LIST<FLOAT NOT NULL>`
```
Hunterness: this should probably be a full sentence and not one sentence and a free hanging part :P

Hunterness (approved):

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -16,6 +16,51 @@ New features are added to the language continuously, and occasionally, some feat
...
+label:functionality[]
+label:updated[]
+[source, cypher, role="noheader"]
+----
+IS [NOT] :: <TYPE>
+----
+a|
+
+Extended xref:syntax/expressions.adoc#type-predicate-expressions[type predicate expressions].
+The newly supported types are:
```
stefano-ottolenghi, line 39:
```suggestion
The newly supported expressions are:
```
Or? At all rates, I don't think it is correct to refer to something like `IS :: STRING`  as a _type_, right?

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1302,36 +1290,55 @@ RETURN val, val IS NOT :: STRING AS notString
 [[type-predicate-expressions-null]]
 === Type predicate expressions for null
 
-All Cypher types include the `null` value.
-Therefore, `IS ::` will return `true` for all expressions evaluating to `null`, regardless of the type.
+A Cypher type will include the `null` value unless it is appended explicitly with `NOT NULL`.
```
stefano-ottolenghi:
```suggestion
A Cypher type includes the `null` value unless it is explicitly appended with `NOT NULL`.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1302,36 +1290,55 @@ RETURN val, val IS NOT :: STRING AS notString
...
+Therefore, `IS ::` will return `true` for all expressions evaluating to `null`, which do not explicitly remove it.
```
stefano-ottolenghi:
```suggestion
Therefore, `IS ::` returns `true` for all expressions evaluating to `null`, unless they do not explicitly remove it.
```

The usage of the explicit `they` clears the ambiguity of whether `which` refers to `null`, as its default in grammar is, or to a previous term `expressions` (as intended in this case).

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1302,36 +1290,55 @@ RETURN val, val IS NOT :: STRING AS notString
...
 
 .Query
 [source, cypher]
 ----
-RETURN (toFloat(null)) IS :: BOOLEAN AS isBoolean
+WITH (toFloat(null)) AS nullValue
```
stefano-ottolenghi:
```suggestion
WITH toFloat(null) AS nullValue
```

I just tested it and the surrounding parenthesis are not needed - unless there is a specific reason to have them? :thinking:

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1302,36 +1290,55 @@ RETURN val, val IS NOT :: STRING AS notString
...
-| isBoolean
-| true
-1+d|Rows: 1
+| isBoolean | isNotNullBoolean
+| true      | false
+2+d|Rows: 1
 |===
 
-Likewise, `IS NOT ::` will return `false` for all expressions evaluating to `null`, regardless of the type.
+Likewise, `IS NOT ::` will return `false` for all expressions evaluating to `null`, if the type is not appended with `NOT NULL`.
```
stefano-ottolenghi:
```suggestion
Likewise, `IS NOT ::` returns `false` for all expressions evaluating to `null`, unless the type is appended with `NOT NULL`.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1302,36 +1290,55 @@ RETURN val, val IS NOT :: STRING AS notString
...
+| false     | true
+2+d|Rows: 1
+|===
+
+It is also possible to check whether a value is only the `null` value using the `NULL` type.
+
+.Query
+[source, cypher]
+----
+WITH (toFloat(null)) AS nullValue
```
stefano-ottolenghi:
```suggestion
WITH toFloat(null) AS nullValue
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1372,6 +1379,9 @@ RETURN n.name AS name, n.age AS age
 2+d|Rows: 1
 |===
 
+The type `PROPERTY VALUE` can also be used to check for whether a type is storable as a property.
```
stefano-ottolenghi:
```suggestion
The type `PROPERTY VALUE` can also be used to check whether a type is storable as a property.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1372,6 +1379,9 @@ RETURN n.name AS name, n.age AS age
...
+Non-storable properties, such as `MAP`, will return `false` when checked with `IS :: PROPERTY VALUE`.
```
stefano-ottolenghi:
```suggestion
Types not storable in properties, such as `MAP`, return `false` when checked with `IS :: PROPERTY VALUE`.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1418,6 +1428,69 @@ For verifying that an expression is not of a certain type, the following alterna
 <expr> IS NOT TYPED <TYPE>
 ----
 
+[[type-predicate-expressions-any-and-nothing]]
+=== Use of Any and Nothing types
```
stefano-ottolenghi:
```suggestion
=== Use of `ANY` and `NOTHING` types
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1418,6 +1428,69 @@ For verifying that an expression is not of a certain type, the following alterna
...
+
+The `ANY` type is a supertype of all types and will therefore match on everything.
```
stefano-ottolenghi:
```suggestion
The `ANY` type is a supertype which matches any type.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1418,6 +1428,69 @@ For verifying that an expression is not of a certain type, the following alterna
...
+The `NOTHING` type is however the type containing an empty set of values. This means that it will return `false` for all values.
```
stefano-ottolenghi:
```suggestion
On the other hand, the `NOTHING` type is the type containing an empty set of values. This means that it returns `false` for all values.
```

stefano-ottolenghi: A curiosity that may potentially result in an extra example here: is there any good use case for `IS :: NOTHING`? When would I want to check that a variable is of _no_ type? How can I even get a variable that is of no type? :thinking:

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
```diff
@@ -88,15 +88,20 @@ Some of these types are not yet supported in either type predicate expressions o
...
 | `POINT` |
-| `RELATIONSHIP` | `EDGE`
+| `PROPERTY VALUE` | `ANY PROPERTY VALUE`
+| `RELATIONSHIP` | `ANY RELATIONSHIP`, `EDGE`, `ANY EDGE`
 | `STRING` | `VARCHAR`
 | `ZONED DATETIME` | `TIMESTAMP WITH TIMEZONE`
 | `ZONED TIME` | `TIME WITH TIMEZONE`
 |===
 
+All Cypher Types contain the `null` value, in order to make them not nullable `NOT NULL` can be appended to the end of the type, e.g `BOOLEAN NOT NULL`, `LIST<FLOAT NOT NULL>`.
```
stefano-ottolenghi:
```suggestion
All Cypher types contain the `null` value. To make them not nullable, `NOT NULL` can be appended to the end of the type (e.g. `BOOLEAN NOT NULL`, `LIST<FLOAT NOT NULL>`).
```

stefano-ottolenghi (commented):
As a general remark, I see examples have `.Query`/`.Result`. This is not a requirement, and just be aware that you may also

- take it away entirely, and the code blocks will lose their bulky blue header
- give it meaningful names, such as `.Check for null type` or whatever

There is no parser expecting the structure with `.Query`, and I also don't think it's of particular value to users if they don't get an informative title :)

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
Hunterness: the type is the `STRING` part, the list below is not repeating the whole expression only the type part of the expression

modules/ROOT/pages/syntax/expressions.adoc
Hunterness: Me and Louise added them to the original line to show it was allowed, it was more useful in the other example with an addition I think even if it wasn't needed there either. Might not be needed in all places using the `toFloat` if there's more than one

modules/ROOT/pages/syntax/expressions.adoc
gem-neo4j: The use case is for LIST<NOTHING>, so an empty list :)

Hunterness: I think the only place you can have something with no type is for empty lists: the inner type of the list is `NOTHING` -> `LIST<NOTHING>`. We don't block the user from writing `IS :: NOTHING` but it'll always be false 🤷

modules/ROOT/pages/syntax/expressions.adoc
gem-neo4j: Saying it this way is not very helpful though? I would prefer not to use the name of the type to describe what it matches on, also just saying supertype doesn't imply it is of all types, so that information is also useful to keep :)

stefano-ottolenghi: > Saying it this way is not very helpful though?

Why not? To a reader, the important bit of information is that _use `ANY` -> match any type_. Then, the fact that the supertype is of all types (and honestly even that it _is_ a supertype at all) is an implementation detail that doesn't further help users. To my eyes, what they care about is that

1. `ANY` is a _type_, so I can use it in any type expression
2. `ANY` matches all types, and acts a bit like a type-wildcard.

If you don't agree, what extra valuable information do you think users get from knowing that it's a _supertype of all types_? Can you come up with a situation where knowing that extra bit allows you to succeed, while not knowing it makes your life harder?

modules/ROOT/pages/syntax/expressions.adoc
stefano-ottolenghi: > is for empty lists: the inner type of the list is NOTHING -> LIST<NOTHING>. 

And when what I would want to do that? What do I do with a `LIST<NOTHING>`? :thinking: 

> We don't block the user from writing IS :: NOTHING but it'll always be false

Is it a meaningful feature then? :thinking:

modules/ROOT/pages/syntax/expressions.adoc
stefano-ottolenghi: I marked it everywhere I saw it in this PR, and I'd recommend taking it away since there is no extra meaning carried by them. I assume developers know that they can put as many surrounding parenthesis as they wish (and this is not specific to Cypher) ... and if they don't know, we don't need to teach them here :D

Hunterness: My initial comment was to add it to the addition to clarify the example (`null + 1 IS NOT :: DATE` is a bit harder to read than `(null + 1) IS NOT :: DATE` even if it means the same thing), it just happened to be added everywhere, I'm fine with removing it from the toFloat examples

modules/ROOT/pages/syntax/expressions.adoc
Hunterness: One thing about the sentence 
```
The `ANY` type is a supertype which matches any type.
```
that I don't particularly like is that it feels like it says nothing by repeating the word `ANY`. 

Maybe something like
```suggestion
The `ANY` type is a supertype which matches values of all types.
```
instead?

modules/ROOT/pages/syntax/expressions.adoc
Hunterness: > What do I do with a `LIST<NOTHING>`?

Confirm if your list is empty or if it has any values in it:
```
[] IS :: LIST<NOTHING> --> true

[null] IS :: LIST<NOTHING> --> false
[1] IS :: LIST<NOTHING> --> false
[true] IS :: LIST<NOTHING> --> false
```

> Is it a meaningful feature then? 🤔

Not to use, but it also feels very odd to block it since it can be an inner type (I guess any examples for it should be with lists but I didn't look that closely on the section to remember what's there), so for consistency it's allowed. (It is also specified as a type in GQL if I remember correctly so it is a general type we should allow the use of.)

modules/ROOT/pages/syntax/expressions.adoc
stefano-ottolenghi: Agreed! I do think it looks nice on math operations.

modules/ROOT/pages/syntax/expressions.adoc
gem-neo4j: Yes, because supertype does not imply all types are below it, it implies at least 1 type is. Number is a supertype of Int and float for example, but it is not a supertype of all types.

It is not generally useful to read a description that uses the same word to describe the word, for example if we used this description for a different type:

The `INTEGER` type is a  supertype which matches integer types.

I have not gained any extra information here. I do not know that Integer matches INT32 and INT64 for example. Describing what a type word means using different words is more useful.

Also technically it doesn't match on a type, it is matching on a value from an expression. New suggestion:

The `ANY` type is a supertype of all Cypher types which evaluates to true for all given expressions.

stefano-ottolenghi: > The `ANY` type is a supertype which matches values of all types.


Yes, I like this!

modules/ROOT/pages/syntax/expressions.adoc
stefano-ottolenghi: > Confirm if your list is empty or if it has any values in it:
[] IS :: LIST<NOTHING> --> true

Yeah, although I find much less opaque checking for `var = []`. Shorter and more clearer what its effect is.

Regardless, thanks for clarifying.

modules/ROOT/pages/syntax/expressions.adoc
stefano-ottolenghi: > The INTEGER type is a supertype which matches integer types.
I have not gained any extra information here. I do not know that Integer matches INT32 and INT64 for example. 

Definitely, but if the supertype is of _all_ types, there is no ambiguity on what types it includes. I do agree the repeated usage of `any` was wrong, but There's suggestion cleared it :)

modules/ROOT/pages/syntax/expressions.adoc
gem-neo4j: It isn't actually specified in GQL, but it is required for the valueType() function that is coming. That function must give the most precise type, for an empty list, according to GQL that is the smallest type. If we don't have `NOTHING`, then the smallest type is either `NULL` or `BOOLEAN`, depending on how we approached it, it was decided that it is confusing to return `LIST<BOOLEAN>` for empty lists, especially in cases where a user is testing a bunch of lists and none of them contain bools. So it is better with `LIST<NOTHING>`.

Another reason is because of list concatenation semantics.

`LIST<NOTHING> + LIST<STRING NOT NULL> = LIST<STRING NOT NULL>`
`[] + ["hi"] = ["hi"] == LIST<STRING NOT NULL>`

if an empty list was `LIST<NULL>` then:
`LIST<NULL> + LIST<STRING NOT NULL> = LIST<STRING>  != LIST<STRING NOT NULL>`
or if an empty list was `LIST<BOOLEAN>` then this is even worse:
`LIST<BOOLEAN> + LIST<STRING NOT NULL> = LIST<BOOLEAN | STRING>  != LIST<STRING NOT NULL>`

So an addition with an empty list would change the type! But as `NOTHING` is a subtype of all types, it plays nice in these cases :D

modules/ROOT/pages/syntax/expressions.adoc
gem-neo4j: This no longer feels natural in English. `Unless` is a negating condition and that alongside the `do not` is hard to read. I think it might even make the sentence incorrect.

Suggestion:

Therefore, `IS ::` returns `true` for all expressions evaluating to `null`, unless they explicitly remove it.

modules/ROOT/pages/syntax/expressions.adoc
Hunterness: > that may potentially result in an extra example

I think it might make sense to add an example of `LIST<NOTHING>` showing an empty list as well as a non-empty list to have a 'useful' example case of `NOTHING`

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1244,19 +1244,7 @@ A type predicate expression can be used to verify the type of a variable, litera
...
-* `DATE`
-* `LOCAL TIME`
-* `ZONED TIME`
-* `LOCAL DATETIME`
-* `ZONED DATETIME`
-* `DURATION`
-* `POINT`
+Where `<expr>` is any Cypher expression and `<TYPE>` is a Cypher Type.
 
 For all available Cypher types, see the section on xref::values-and-types/property-structural-constructed.adoc#types-synonyms[types and their synonyms].
```
Hunterness:
```suggestion
Where `<expr>` is any Cypher expression and `<TYPE>` is a Cypher Type.
For all available Cypher types, see the section on xref::values-and-types/property-structural-constructed.adoc#types-synonyms[types and their synonyms].
```
I'd probably make them one paragraph instead of two 🤷

modules/ROOT/pages/syntax/expressions.adoc
stefano-ottolenghi: You are right, the `not` slipped there from the initial sentence. Do take it away :)

modules/ROOT/pages/syntax/expressions.adoc
gem-neo4j: This is already there :) Under the list types part (the next section)

stefano-ottolenghi (approved):

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1302,36 +1215,49 @@ RETURN val, val IS NOT :: STRING AS notString
...
+
+[role="queryresult",options="header,footer",cols="2*<m"]
+|===
+| isNotDate | isNotNotNullDate
+| false     | true
+2+d|Rows: 1
 |===
 
-Likewise, `IS NOT ::` will return `false` for all expressions evaluating to `null`, regardless of the type.
+It is also possible to check whether a value is only the `null` value using the `NULL` type.
```
JPryce-Aklundh: Maybe this changes meaning of sentence - but I couldnt understand the meaning as it is currently. 

```suggestion
It is also possible to check whether a value is the only `null` value using the `NULL` type.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1418,6 +1343,63 @@ For verifying that an expression is not of a certain type, the following alterna
 <expr> IS NOT TYPED <TYPE>
 ----
 
+[[type-predicate-expressions-any-and-nothing]]
+=== Use of `ANY` and `NOTHING` types
+
+The `ANY` type is a supertype which matches values of all types.
+On the other hand, the `NOTHING` type is the type containing an empty set of values. This means that it returns `false` for all values.
```
JPryce-Aklundh:
```suggestion
`ANY` is a supertype which matches values of all types.
`NOTHING` is a type containing an empty set of values.
This means that it returns `false` for all values.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1418,6 +1343,63 @@ For verifying that an expression is not of a certain type, the following alterna
...
+|===
+| isOfTypeAny | isOfTypeNothing
+| true        | false
+2+d|Rows: 1
+|===
+
+[[type-predicate-expressions-lists]]
+=== List Types
+
+Type predicate expressions can be used for `LIST` types, where the inner type of the list must be specified.
```
JPryce-Aklundh: Maybe 'elements' is not correct here, but I think it clarifies the below mentions of "inner type"

```suggestion
Type predicate expressions can be used for `LIST` types, where the inner type of the elements in the list must be specified.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1302,36 +1215,49 @@ RETURN val, val IS NOT :: STRING AS notString
 [[type-predicate-expressions-null]]
 === Type predicate expressions for null
 
-All Cypher types include the `null` value.
-Therefore, `IS ::` will return `true` for all expressions evaluating to `null`, regardless of the type.
+A Cypher type includes the `null` value unless it is explicitly appended with `NOT NULL`.
```
JPryce-Aklundh:
```suggestion
As of Neo4j 5.10, a Cypher type includes the `null` value unless it is explicitly appended with `NOT NULL`.
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1235,7 +1165,7 @@ RETURN COLLECT {
 [[type-predicate-expressions]]
 == Type predicate expressions
 
-_This feature was introduced in Neo4j 5.9._
+_This feature was introduced in Neo4j 5.9 and extended in 5.10_
```
JPryce-Aklundh:
```suggestion
_This feature was introduced in Neo4j 5.9._
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1418,6 +1343,63 @@ For verifying that an expression is not of a certain type, the following alterna
 <expr> IS NOT TYPED <TYPE>
 ----
 
+[[type-predicate-expressions-any-and-nothing]]
+=== Use of `ANY` and `NOTHING` types
+
```
JPryce-Aklundh, line 1348:
```suggestion
=== Use of `ANY` and `NOTHING` types

_This feature was introduced in Neo4j 5.10._
```

modules/ROOT/pages/syntax/expressions.adoc
```diff
@@ -1418,6 +1343,63 @@ For verifying that an expression is not of a certain type, the following alterna
...
+[role="queryresult",options="header,footer",cols="2*<m"]
+|===
+| isOfTypeAny | isOfTypeNothing
+| true        | false
+2+d|Rows: 1
+|===
+
+[[type-predicate-expressions-lists]]
+=== List Types
+
```
JPryce-Aklundh, line 1369:
```suggestion
=== List Types

_This feature was introduced in 5.10._
```

JPryce-Aklundh (commented):
Thanks @gem-neo4j  - I have made a few suggestions to make more clear what is available with the release of 5.10, as well as a few other minor suggestions.

Also, with regard to Query/Result headers - I have less against them than @stefano-ottolenghi as I think, if used consistently, they add to the readability of the Manual. But, that is a docs-issue to resolve (and whatever we decide, you don't need to do anything else with them in this PR).

JPryce-Aklundh (approved):

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
```diff
@@ -88,15 +88,20 @@ Some of these types are not yet supported in either type predicate expressions o
...
 | `POINT` |
-| `RELATIONSHIP` | `EDGE`
+| `PROPERTY VALUE` | `ANY PROPERTY VALUE`
+| `RELATIONSHIP` | `ANY RELATIONSHIP`, `EDGE`, `ANY EDGE`
 | `STRING` | `VARCHAR`
 | `ZONED DATETIME` | `TIMESTAMP WITH TIMEZONE`
 | `ZONED TIME` | `TIME WITH TIMEZONE`
 |===
 
+All Cypher types contain the `null` value. To make them not nullable, `NOT NULL` can be appended to the end of the type (e.g. `BOOLEAN NOT NULL`, `LIST<FLOAT NOT NULL>`).
```
ncordon, line 103: Maybe specify except for `NULL`? I suppose `NULL NOT NULL` is not valid?

ncordon (approved):

modules/ROOT/pages/values-and-types/property-structural-constructed.adoc
gem-neo4j: It is valid actually, NULL NOT NULL == NOTHING and NOTHING NOT NULL == NOTHING as well :D

PR #653, opened 2023-07-11
Add a note to SHOW ALIASES as we now filter the database column

mnd999:
The database column is now filtered in the same was as SHOW DATABASES, so we add a note to the documentation to explain that.


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-653.surge.sh

Hunterness (approved):

modules/ROOT/pages/administration/aliases.adoc
```diff
@@ -283,6 +283,18 @@ The required privileges are described xref::administration/access-control/dbms-a
...
 | database
 | The name of the target database. label:default-output[]
+
+This column is filtered according to the `ACCESS` privileges of the user.
+However, some privileges enable users to see additional databases regardless of their `ACCESS` privileges:
+
+* Users with `CREATE/DROP/ALTER DATABASE` or `SET DATABASE ACCESS` privileges can see all standard databases.
+* Users with `CREATE/DROP COMPOSITE DATABASE` or `COMPOSITE DATABASE MANAGEMENT` privileges can see all composite databases.
+* Users with `DATABASE MANAGEMENT` privilege can see all databases.
+* Users always have visibility of the `system` database.
```
JPryce-Aklundh:
```suggestion
* Users can always see the `system` database.
```

modules/ROOT/pages/administration/aliases.adoc
```diff
@@ -283,6 +283,18 @@ The required privileges are described xref::administration/access-control/dbms-a
 
 | database
 | The name of the target database. label:default-output[]
+
+This column is filtered according to the `ACCESS` privileges of the user.
+However, some privileges enable users to see additional databases regardless of their `ACCESS` privileges:
+
+* Users with `CREATE/DROP/ALTER DATABASE` or `SET DATABASE ACCESS` privileges can see all standard databases.
```
JPryce-Aklundh:
```suggestion
* Users with the `CREATE DATABASE`, `DROP DATABASE`, `ALTER DATABASE`, or `SET DATABASE ACCESS` privileges can see all standard databases.
```

modules/ROOT/pages/administration/aliases.adoc
```diff
@@ -283,6 +283,18 @@ The required privileges are described xref::administration/access-control/dbms-a
...
+* Users with `CREATE/DROP COMPOSITE DATABASE` or `COMPOSITE DATABASE MANAGEMENT` privileges can see all composite databases.
```
JPryce-Aklundh:
```suggestion
* Users with the `CREATE COMPOSITE DATABASE`, `DROP COMPOSITE DATABASE`, or `COMPOSITE DATABASE MANAGEMENT` privileges can see all composite databases.
```

modules/ROOT/pages/administration/aliases.adoc
```diff
@@ -283,6 +283,18 @@ The required privileges are described xref::administration/access-control/dbms-a
...
+* Users with `DATABASE MANAGEMENT` privilege can see all databases.
```
JPryce-Aklundh:
```suggestion
* Users with the `DATABASE MANAGEMENT` privilege can see all databases.
```

modules/ROOT/pages/administration/aliases.adoc
```diff
@@ -283,6 +283,18 @@ The required privileges are described xref::administration/access-control/dbms-a
...
+* Users always have visibility of the `system` database.
+
+If a user has not been granted `ACCESS` privilege nor any of the above special cases above to the target database, then it is not visible to the user
```
JPryce-Aklundh:
```suggestion
If a user has not been granted any of these privileges to the target database, then it is not visible and this column will be `null`. 
```

modules/ROOT/pages/administration/aliases.adoc
```diff
@@ -283,6 +283,18 @@ The required privileges are described xref::administration/access-control/dbms-a
...
+and this column will be `null`.
```
JPryce-Aklundh, line 296:
```suggestion
```

modules/ROOT/pages/administration/aliases.adoc
Hunterness: If we change this here we should probably change it in the database chapter as well as it has the same text

modules/ROOT/pages/administration/aliases.adoc
mnd999: This changes the meaning somewhat. The above list of privileges are not database specific, they appy to the DBMS. And we need to mention the ACCESS privilege really as that can apply to specific databases and it is not mentioned in the list above.

modules/ROOT/pages/administration/aliases.adoc
```diff
@@ -283,6 +283,18 @@ The required privileges are described xref::administration/access-control/dbms-a
...
+
+This column is filtered according to the `ACCESS` privileges of the user.
+However, some privileges enable users to see additional databases regardless of their `ACCESS` privileges:
+
+* Users with the `CREATE DATABASE`, `DROP DATABASE`, `ALTER DATABASE`, or `SET DATABASE ACCESS` privileges can see all standard databases.
+* Users with the `CREATE COMPOSITE DATABASE`, `DROP COMPOSITE DATABASE`, or `COMPOSITE DATABASE MANAGEMENT` privileges can see all composite databases.
+* Users with the `DATABASE MANAGEMENT` privilege can see all databases.
+* Users can always see the `system` database.
+
+If a user has not been granted `ACCESS` privilege nor any of the above special cases above to the target database, then it is not visible to the user
```
mnd999: How about this?
```suggestion
If a user has not been granted the `ACCESS` privilege to the target database and none of the above special cases apply, then it is not visible to the user
```

modules/ROOT/pages/administration/aliases.adoc
JPryce-Aklundh: Ok - I will delete the second "above" in "above special cases above" and then commit.

modules/ROOT/pages/administration/aliases.adoc
JPryce-Aklundh: Great 👍

JPryce-Aklundh (approved):

PR #664, opened 2023-07-13
Remove shortest var length and shortest path examples from MATCH page 

JPryce-Aklundh:
This information is now in the Patterns chapter. 

Trello:
https://trello.com/c/dOwcPco6/5144-remove-shortestpath-and-var-length-examples-from-match-page


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-cypher-664.surge.sh

PR #670, opened 2023-08-01
Restructure the main part of create.adoc

arnefischereit:
... in order to be able to talk about the clauses restrictions.

### DIFF
--- a/modules/ROOT/pages/clauses/create.adoc
+++ b/modules/ROOT/pages/clauses/create.adoc
@@ -6,17 +6,16 @@
 == Introduction
 
 The `CREATE` clause allows you to create nodes and relationships.
-Similar to xref::clauses/match.adoc[`MATCH`], this is done using xref::patterns/index.adoc[patterns].
-However, the patterns you can use in `CREATE` are restricted to what is relevant to describe the newly created nodes and relationships.
+To define these entities, `CREATE` uses a syntax similar to that of xref::clauses/match.adoc[`MATCH`].
+However, while xref::patterns/index.adoc[patterns] only need to evaluate to either true or false, the syntax for `CREATE` needs to specify exactly what nodes and relationships to create.
 
 [[create-nodes]]
 == Syntax for nodes
 
 The `CREATE` clause allows you to create one or more nodes.
-For each, you can specify labels and properties.
-You can give the nodes names that you can refer to later in the query.
-xref:clauses/where.adoc[`WHERE`-subclauses] and xref:clauses/match.adoc#label-expression-match-or-expression[label expression] are not allowed.
-Instead, multiple labels are separated by colons.
+Each node can be assigned labels and properties.
+You can bind each node to a variable that you can refer to later in the query.
+Multiple labels are separated by colons.
 
 .Query
 [source, cypher, indent=0]
@@ -24,26 +23,15 @@ Instead, multiple labels are separated by colons.
 CREATE (charlie:Person:Actor {name: 'Charlie Sheen'}), (oliver:Person:Director {name: 'Oliver Stone'})
 ----
 
-Creates two nodes, each with a `name` and a label `Person`.
-The one representing Charlie Sheen also has the label `Actor` while the one representing Oliver Stone receives the label `Director`.
-
-.Result
-[role="queryresult",options="footer",cols="1*<m"]
-|===
-1+|(empty result)
-1+d|Rows: 0 +
-Nodes created: 2 +
-Properties set: 2 +
-Labels added: 4
-|===
+Creates two nodes, bound to the variables `charlie` and `oliver`, each with a `Person` label and a `name` property.
+The node representing Charlie Sheen also has the label `Actor` while the node representing Oliver Stone is assigned the label `Director`.
 
 [[create-relationships]]
 == Syntax for relationships
 
-You can create relationships very similarly to nodes.
-However, relationships always need exactly one relationship type and a direction.
-Beyond this, they may also have properties and names.
-They may not be quantified or carry `WHERE`-subclause.
+Relationships can also be created using the `CREATE` clause.
+Unlike nodes, relationships always need exactly one relationship type and a direction.
+Similar to nodes, relationships can be assigned properties and relationship types and be bound to variables.
 
 .Query
 [source, cypher, indent=0]
@@ -51,24 +39,15 @@ They may not be quantified or carry `WHERE`-subclause.
 CREATE (charlie:Person:Actor {name: 'Charlie Sheen'})-[:ACTED_IN {role: 'Bud Fox'}]->(wallStreet:Movie {title: 'Wall Street'})<-[:DIRECTED]-(oliver:Person:Director {name: 'Oliver Stone'})
 ----
 
-Creates not only the nodes for Charlie Sheen and Oliver Stone as well as the one for the movie Wall Street but also the relationships of type `ACTED_IN` and `DIRECTED` between them.
-
-.Result
-[role="queryresult",options="footer",cols="1*<m"]
-|===
-1+|(empty result)
-1+d|Rows: 0 +
-Nodes created: 3 +
-Relationships created: 2 +
-Properties set: 5 +
-Labels added: 5
-|===
+This query creates the `Person` nodes for Charlie Sheen and Oliver Stone and the `Movie` node for Wall Street.
+It also created the relationships of the types `ACTED_IN` and `DIRECTED` between them.
 
 [[reusing-variables]]
 == Reusing variables
-The previous example created a whole path at once.
-Note, that these newly created elements are now disconnected from the rest of the graph.
-To instead connect them to already existing data, you can refer to variables from other parts of the query.
+The previous example created a path between the specified nodes.
+Note, that these newly created nodes and relationships are not connected to what was previously in the graph.
+To connect them to already existing data, bind the desired nodes and relationships to variables.
+These variables can then be passed along to subsequent clauses in a query that target pre-existing elements in the graph.
 
 .Query
 [source, cypher, indent=0]
@@ -77,7 +56,8 @@ MATCH (charlie:Person {name: 'Charlie Sheen'}), (oliver:Person {name: 'Oliver St
 CREATE (charlie)-[:ACTED_IN {role: 'Bud Fox'}]->(wallStreet:Movie {title: 'Wall Street'})<-[:DIRECTED]-(oliver)
 ----
 
-First looks up the two nodes representing Charlie Sheen and Oliver Stone in order to then connect them in the same way as we have done in the previous query.
+In this example, the `MATCH` clause finds the nodes Charlie Sheen and Oliver Stone and binds them to the `charlie` and `oliver` variables respectively.
+These variables are then passed along to the subsequent `CREATE` clause, which creates new relationships from the bound nodes.
 
 You can also reuse variables from the same `CREATE`, both in the same or a later clause.
 This way, you can, for example, define constructs that are more complex than just a linear path.
@@ -92,7 +72,7 @@ RETURN length(p)
 Creates all three nodes for Charlie Sheen, Oliver Stone and Michael Douglas and connects them all to the node representing the Wall Street movie.
 It then returns the length of the path from Charlie Sheen to Oliver Stone.
 
-Note that when repeating a node's name, you may not apply labels or properties to the repetition.
+Note that when repeating a node's variable, you may not add labels or properties to the repetition.
 
 .Query
 [source, cypher, role=test-fail]
@@ -101,7 +81,7 @@ MATCH (charlie:Person {name: 'Charlie Sheen'})
 CREATE (charlie:Actor)
 ----
 
-This query will fail because the node `charlie` already exists, and we therefore cannot `CREATE` it.
+This query will fail because the variable `charlie` has already been bound to a pre-existing node, and therefore it cannot be reused to create a new node.
 If you intend to add a label, use the xref:clauses/set.adoc#set-set-a-label-on-a-node[`SET` clause] instead.
 
 [[reusing-variables-properties]]
@@ -126,10 +106,10 @@ MATCH (person:Person)
 CREATE (anotherPerson:Person {name: person.name, age: $age})
 ----
 
-Creates a person with the same name as another person and the age from a xref:syntax/parameters.adoc[parameter] called `age`.
+This example created a `Person` node with the same name as another person and the age from a xref:syntax/parameters.adoc[parameter] called `age`.
 
 Such an expression may not contain a reference to a variable that is defined in the same `CREATE` statement.
-This is defined this way so that it is always clear what value a property should have.
+This is to ensure that the value of a property is always clear.
 
 .Query
 [source, cypher, role=test-fail]
@@ -137,8 +117,8 @@ This is defined this way so that it is always clear what value a property should
 CREATE (charlie {score: oliver.score + 1}), (oliver {score: charlie.score + 1})
 ----
 
-Tries to create nodes such that Charlie's score is higher than Oliver's and vice versa, which is a contradiction.
-Therefore, queries like this are rejected.
+This query tries to create nodes such that Charlie's score is higher than Oliver's and vice versa, which is a contradiction.
+The query therefore fails.
 
 [[use-parameters-with-create]]
 == Use parameters with `CREATE`
--- a/modules/ROOT/pages/clauses/create.adoc
+++ b/modules/ROOT/pages/clauses/create.adoc
@@ -3,247 +3,142 @@
 [[query-create]]
 = CREATE
 
-[abstract]
---
-The `CREATE` clause is used to create nodes and relationships.
---
+== Introduction
 
+The `CREATE` clause allows you to create nodes and relationships.
+Similar to xref::clauses/match.adoc[`MATCH`], this is done using xref::patterns/index.adoc[patterns].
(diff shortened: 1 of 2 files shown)


neo-technology-commit-status-publisher:

arnefischereit:
Direct link to CREATE page:
https://neo4j-docs-cypher-670.surge.sh/cypher-manual/5/clauses/create/

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
 
-[abstract]
---
-The `CREATE` clause is used to create nodes and relationships.
---
+== Introduction
 
+The `CREATE` clause allows you to create nodes and relationships.
+Similar to xref::clauses/match.adoc[`MATCH`], this is done using xref::patterns/index.adoc[patterns].
+However, the patterns you can use in `CREATE` are restricted to what is relevant to describe the newly created nodes and relationships.
```
blouerat: I feel uneasy about this part.
You create elements by spelling out explicit nodes or paths: `(:A)`, `(:A)-[r:R]->({foo: "bar"})`.
You match on elements using nodes or path patterns. A path pattern represents the set of paths in the graph matching the pattern at the time of execution of the query.
The node `(:A)` is very different indeed from the node pattern `(:A)`, the lattern will also return node `(:A:B)` for example.
The syntax for a path happens to be valid syntax for a path pattern, but I'm not sure it is useful to present the path syntax as a restricted version of the path pattern syntax.
The same way the string "hello" is a valid regex that will match on the string "hello". When you ask the user for a string, you don't tell them to enter a restricted regular expression.
I know that we haven't made that distinction in our parser or in our AST[^1], but I think we should make it clearer to our users.

[^1]: I think we should differentiate it there as well but it's a different conversation.

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
 
-[[create-create-single-node]]
-=== Create single node
-
-Creating a single node is done by issuing the following query:
+The `CREATE` clause allows you to create one or more nodes.
+For each, you can specify labels and properties.
+You can give the nodes names that you can refer to later in the query.
+xref:clauses/where.adoc[`WHERE`-subclauses] and xref:clauses/match.adoc#label-expression-match-or-expression[label expression] are not allowed.
+Instead, multiple labels are separated by colons.
```
blouerat: This follows from my initial comment, `WHERE` only makes sense in a node pattern, not in a node.

arnefischereit:
@blouerat and I concluded that we need to find a common stance in the company on whether we should see `CREATE` and `MATCH` to share (parts of) the patterns they accept.

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
-== Create nodes
+== Syntax for nodes
 
-[[create-create-single-node]]
-=== Create single node
-
-Creating a single node is done by issuing the following query:
+The `CREATE` clause allows you to create one or more nodes.
+For each, you can specify labels and properties.
+You can give the nodes names that you can refer to later in the query.
```
JPryce-Aklundh: When we say node "names" - we just mean labels, right? If so, I think it would be better to just stick with labels, as right now the 2 could be interpreted to mean to different things.

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
 [[create-nodes]]
-== Create nodes
+== Syntax for nodes
 
-[[create-create-single-node]]
-=== Create single node
-
-Creating a single node is done by issuing the following query:
+The `CREATE` clause allows you to create one or more nodes.
+For each, you can specify labels and properties.
```
JPryce-Aklundh:
```suggestion
Each node can be assigned labels and properties.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
-=== Create multiple nodes
-
-Creating multiple nodes is done by separating them with a comma.
-
-.Query
-[source, cypher, indent=0]
-----
-CREATE (n), (m)
-----
+Creates two nodes, each with a `name` and a label `Person`.
```
JPryce-Aklundh:
```suggestion
Creates two nodes, each with a `Person` label and a `name` property.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
+The one representing Charlie Sheen also has the label `Actor` while the one representing Oliver Stone receives the label `Director`.
```
JPryce-Aklundh:
```suggestion
The node representing Charlie Sheen also has the label `Actor` while the node representing Oliver Stone is assigned the label `Director`.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
 .Result
 [role="queryresult",options="footer",cols="1*<m"]
 |===
 1+|(empty result)
 1+d|Rows: 0 +
-Nodes created: 2
+Nodes created: 2 +
+Properties set: 2 +
+Labels added: 4
 |===
```
JPryce-Aklundh: I think this can be removed.

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
 
+[[create-relationships]]
+== Syntax for relationships
 
-[[create-create-a-node-with-a-label]]
-=== Create a node with a label
-
-To add a label when creating a node, use the syntax below:
+You can create relationships very similarly to nodes.
```
JPryce-Aklundh:
```suggestion
Relationships can also be created using the `CREATE` clause.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
+However, relationships always need exactly one relationship type and a direction.
```
JPryce-Aklundh:
```suggestion
Unlike nodes, relationships always need exactly one relationship type and a direction.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
+Beyond this, they may also have properties and names.
+They may not be quantified or carry `WHERE`-subclause.
```
JPryce-Aklundh: Again, I am unsure about the saying relationships have "names"? Maybe I am missing something?

Also, maybe we should elaborate on what quantified means in this context? I am guessing you mean you need to create them one by one?

```suggestion
Similar to nodes, relationships can be assigned properties and names.
They may not be quantified or carry a `WHERE`-subclause.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
 
 .Query
 [source, cypher, indent=0]
 ----
-CREATE (n:Person)
+CREATE (charlie:Person:Actor {name: 'Charlie Sheen'})-[:ACTED_IN {role: 'Bud Fox'}]->(wallStreet:Movie {title: 'Wall Street'})<-[:DIRECTED]-(oliver:Person:Director {name: 'Oliver Stone'})
 ----
 
+Creates not only the nodes for Charlie Sheen and Oliver Stone as well as the one for the movie Wall Street but also the relationships of type `ACTED_IN` and `DIRECTED` between them.
```
JPryce-Aklundh:
```suggestion
This query creates the `Person` nodes for Charlie Sheen and Oliver Stone and the `Movie` node for Wall Street.
It also created the relationships of the types `ACTED_IN` and `DIRECTED` between them.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
 [role="queryresult",options="footer",cols="1*<m"]
 |===
 1+|(empty result)
 1+d|Rows: 0 +
-Nodes created: 1 +
-Labels added: 1
+Nodes created: 3 +
+Relationships created: 2 +
+Properties set: 5 +
+Labels added: 5
```
JPryce-Aklundh: Same as before, I think this could be removed.

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
-
-[[create-create-a-node-with-multiple-labels]]
-=== Create a node with multiple labels
-
-To add labels when creating a node, use the syntax below.
-In this case, we add two labels.
+[[reusing-variables]]
+== Reusing variables
+The previous example created a whole path at once.
+Note, that these newly created elements are now disconnected from the rest of the graph.
```
JPryce-Aklundh: I am unsure about what you mean with the word "now" in the second sentence?
Do you mean that these newly created nodes and relationships are not connected to what was previously in the graph?

```suggestion
The previous example created a path between the specified nodes.
Note, that these newly created nodes and relationships are not connected to what was previously in the graph.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
+To instead connect them to already existing data, you can refer to variables from other parts of the query.
```
JPryce-Aklundh: Is this a correct interpretation?

```suggestion
To connect them to already existing data, bind the desired nodes and relationships to variables.
These variables can then be passed along to subsequent clauses in a query that target pre-existing elements in the graph.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
-1+|(empty result)
-1+d|Rows: 0 +
-Nodes created: 1 +
-Labels added: 2
-|===
-
-
-[[create-create-node-and-add-labels-and-properties]]
-=== Create node and add labels and properties
+First looks up the two nodes representing Charlie Sheen and Oliver Stone in order to then connect them in the same way as we have done in the previous query.
```
JPryce-Aklundh:
```suggestion
In this example, the `MATCH` clause finds the nodes Charlie Sheen and Oliver Stone and binds them to the `charlie` and `oliver` variables respectively. 
These variables are then passed along to the subsequent `CREATE` clause, which creates new relationships from the bound nodes.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
-|===
-
+Creates all three nodes for Charlie Sheen, Oliver Stone and Michael Douglas and connects them all to the node representing the Wall Street movie.
+It then returns the length of the path from Charlie Sheen to Oliver Stone.
 
-[[create-return-created-node]]
-=== Return created node
-
-Creating a single node is done by issuing the following query:
+Note that when repeating a node's name, you may not apply labels or properties to the repetition.
```
JPryce-Aklundh: Now I am a little bit more confused about what a node's name is? In this context it seems to suggest it is not a label, but a variable?

```suggestion
Note that when repeating a node's name, you may not add labels or properties to the repetition.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
-.Result
-[role="queryresult",options="header,footer",cols="1*<m"]
-|===
-| +a.name+
-| +"Andy"+
-1+d|Rows: 1 +
-Nodes created: 1 +
-Properties set: 1
-|===
+This query will fail because the node `charlie` already exists, and we therefore cannot `CREATE` it.
```
JPryce-Aklundh: I think this makes it clearer:
```suggestion
This query will fail because the variable `charlie` has already been bound to a pre-existing node, and therefore it cannot be reused to create a new node.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
-| +"RELTYPE"+ | +"A<->B"+
-2+d|Rows: 1 +
-Relationships created: 1 +
-Properties set: 1
-|===
-
-
-[[create-create-a-full-path]]
-== Create a full path
+Creates a person with the same name as another person and the age from a xref:syntax/parameters.adoc[parameter] called `age`.
```
JPryce-Aklundh:
```suggestion
This example created a `Person` node with the same name as another person and the age from a xref:syntax/parameters.adoc[parameter] called `age`.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
 
-When you use `CREATE` and a pattern, all parts of the pattern that are not already in scope at this time will be created.
+Such an expression may not contain a reference to a variable that is defined in the same `CREATE` statement.
+This is defined this way so that it is always clear what value a property should have.
```
JPryce-Aklundh:
```suggestion
This is to ensure that the value of a property is always clear.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
 
 .Query
-[source, cypher]
+[source, cypher, role=test-fail]
 ----
-CREATE p = (:Person {name:'Andy'})-[:WORKS_AT]->(:Company {name: 'Neo4j'})<-[:WORKS_AT]-(:Person {name: 'Michael'})
-RETURN p
+CREATE (charlie {score: oliver.score + 1}), (oliver {score: charlie.score + 1})
```
JPryce-Aklundh, line 117:
```suggestion
MATCH (charlie:Person {name: 'Charlie Sheen), (oliver:Person {name: 'Oliver Stone'})
CREATE (charlie {score: oliver.score + 1}), (oliver {score: charlie.score + 1})
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
-|===
-| +p+
-| (:Person {name: "Andy"})-[:WORKS_AT]->(:Company {name: "Neo4j"})<-[:WORKS_AT]-(:Person {name: "Michael"})
-1+d|Rows: 1 +
-Nodes created: 3 +
-Relationships created: 2 +
-Properties set: 2
-|===
-
+Tries to create nodes such that Charlie's score is higher than Oliver's and vice versa, which is a contradiction.
```
JPryce-Aklundh:
```suggestion
This query tries to create nodes such that Charlie's score is higher than Oliver's and vice versa, which is a contradiction.
The query therefore fails.
```

modules/ROOT/pages/clauses/create.adoc
```diff
@@ -3,247 +3,142 @@
...
+Therefore, queries like this are rejected.
```
JPryce-Aklundh:
```suggestion
```

JPryce-Aklundh (commented):
Thanks, @arnefischereit - I have made some comments to clarify what I thought was potentially ambiguous. Let me know what you think.
Also, as stated in a private conversation, I agree with @blouerat that we should more carefully distinguish between a node and a node pattern.

modules/ROOT/pages/clauses/create.adoc
arnefischereit: I mean identifiers/variables, as in this case `charlie` and `oliver`. Maybe I should call it "identifier" or "variable" then.

modules/ROOT/pages/clauses/create.adoc
arnefischereit: Yes, that captures it. "Now" was meant like "after the `CREATE` took effect".

modules/ROOT/pages/clauses/create.adoc
arnefischereit: I don't mind but I tried to copy the style used in the `MATCH` page, because I thought that was the current best practice.

There, we for example have:

> By specifying a pattern with a single node and no labels, all nodes in the graph will be returned.
> Query
> 
> MATCH (n)
> RETURN n
> 
> Returns all the nodes in the database.

Now, I am just a bit unsure whether that is the preferred style.

modules/ROOT/pages/clauses/create.adoc
JPryce-Aklundh: I think "variable" is preferable.

modules/ROOT/pages/clauses/create.adoc
arnefischereit: No, that would fail for different reasons.

modules/ROOT/pages/clauses/create.adoc
arnefischereit: I removed the sentence on quantified relationships and `WHERE`-clauses as it only made sense when defining a pattern to follow a `CREATE`.

modules/ROOT/pages/clauses/create.adoc
arnefischereit: I added:

> You can bind each node to a variable that you can refer to later in the query.

arnefischereit:
I rephrased the first paragraph, copied most of @JPryce-Aklundh's suggestions and commented on the others.

I hope you both agree, @blouerat, @JPryce-Aklundh.

blouerat (approved):

JPryce-Aklundh (approved):

modules/ROOT/pages/clauses/create.adoc
JPryce-Aklundh: I see what you mean, and re-reading I think both works.

PR #675, opened 2023-08-02
UNION combines `any` queries

aksanoble:
The UNION clause combines results that match `any` of the queries and not `all` queries. The current explanation seems to be incorrect.

PR #683, opened 2023-08-09
UNION combines `any` queries (#675)

JPryce-Aklundh:
The UNION clause combines results that match `any` of the queries and not `all` queries. The current explanation seems to be incorrect.


neo-technology-commit-status-publisher:

PR #690, opened 2023-08-14
Document vector indexes

parnmatt:


parnmatt:
Related changes in operations manual: https://github.com/neo4j/docs-operations/pull/970

renetapopova:
@recrwplay, same here, can't build it locally because of this missing module.

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,421 @@
...
+
+.+DROP INDEX+
+======
+
+In the following example, we will drop the `abstract-embeddings` that we created previously:
+
+.Query
+[source,cypher]
+----
+DROP INDEX abstract-embeddings
```
nilsceberg: I'm pretty sure this would need backticks due to the dash, so maybe an unfortunate example name :P

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,421 @@
...
+Cosine interprets the vectors in Cartesian coordinates.
+The measure is related to the angle between the two vectors; however, an angle can be described in many units, sign conventions, and periods.
+The trigonometric cosine of this angle is both agnostic to the aforementioned and bounded.
+Cosine similar rebounds the trigonometric cosine.
+
+[latexmath]
+++++
+\begin{align}
+    \mathrm{cosine}(\vec v,\vec u)
+        &:= \tfrac12\left(1 + \hat v\cdot\hat u\right) \\[1ex]
```
nilsceberg: Again I think it would be nice to clarify the unit vector notation

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,421 @@
...
+
+A valid vector for a Cosine vector index is when:
+
+* all vector components can be represented finitely in {ieee-754} single precision,
+* and its latexmath:[\ell^2]-norm is non-zero and can be represented finitely in {ieee-754} single precision.
+
+Cosine interprets the vectors in Cartesian coordinates.
+The measure is related to the angle between the two vectors; however, an angle can be described in many units, sign conventions, and periods.
+The trigonometric cosine of this angle is both agnostic to the aforementioned and bounded.
+Cosine similar rebounds the trigonometric cosine.
```
nilsceberg: Similarity?

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,421 @@
...
+----
+(no changes, no records)
+----
+
+We can see that the vector index has been created using `SHOW INDEXES`
+
+.Query
+[source,cypher]
+----
+SHOW INDEXES YIELD name, state, type, labelsOrTypes, properties, OPTIONS
```
nilsceberg: Shoulds options be lowercase here (it's not a keyword in this context)?

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,421 @@
...
+
+.Query
+[source,cypher]
+----
+MATCH (title:Title)<--(:Paper)-->(abstract:Abstract)
+WHERE toLower(title.text) = 'efficient and robust approximate nearest neighbor search using
+  hierarchical navigable small world graphs'
+
+CALL db.index.vector.queryNodes('abstract-embeddings', 10, abstract.embedding)
+YIELD NODE AS similarAbstract, score
```
nilsceberg: Lowercase `node`?

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,421 @@
...
+For instance, assume we have a graph of research papers, and each paper has an abstract; we wish to find papers in the neighborhood of a paper we know.
+
+.Data model
+[source,cypher]
+----
+(:Title)<--(:Paper)-->(:Abstract)
+----
+
+Assume for each abstract, we have generated a 1536-dimensional vector `embedding` of the abstract's `text` using Open AI's default model, `text-embedding-ada-002`.
+This model suggestions a xref:indexes-vector-similarity-cosine[cosine similarity].
```
nilsceberg: suggests :)

nilsceberg:
I've had a look mostly from a technical point of view and I think this looks great, nice job :+1:

neo-technology-commit-status-publisher:

renetapopova:
Thanks, @nilsceberg, I'll then review it from an editorial perspective 😄 .

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: I don't know why, byt it kept making it `OPTIONS`, I swear I changed that!!

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: likewise here … what the heck!

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: IntelliJ autoformat, must have tied into the cypher inline syntax and tried to help out :roll_eyes:

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,436 @@
...
+The index name must be unique among both indexes and constraints.
+====
+
+A vector index is a single-label, single-property index for nodes.
+In addition to the label and property key, both given as `STRING`, a vector index needs to be configured with both the dimensionality of the vector, `INTEGER`, and the measure of similarity between two vectors, case-insensitive `STRING` For details, see xref:#indexes-vector-similarity[].
+
+.Syntax for creating vector indexes
+[source,cypher]
+----
+CALL db.index.vector.createNodeIndex('index_name', 'LabelName', 'propertyKey', dimensionality, 'similarity_function')
```
renetapopova: The test for this code is failing because it thinks `dimensionality` is a variable.

parnmatt: … ok, how interesting … does the test try and execute cypher?
eitherway I changed this to just be the signature, as was done by you later on.

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+
+[[indexes-vector]]
+= Nearest neighbor search
+
+[abstract]
+--
+This chapter describes how to use vector indexes to perform an approximate nearest neighbor search.
+--
+
+_This feature is introduced in Neo4j 5.11._ label:beta[]
```
JPryce-Aklundh:
```suggestion
_This feature was introduced in Neo4j 5.11._ label:beta[]

This chapter describes how to use vector indexes to perform an approximate nearest neighbor search.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+
+Vector indexes are powered by the link:https://lucene.apache.org/[Apache Lucene] indexing and search library, and can be used to index nodes by `LIST<FLOAT>` properties valid to the index.
+A vector index allows you to write queries that match a _neighborhood_ of nodes based on the similarity of those node properties to the query.
+Lucene implements a Hierarchical Navigable Small World{fn-hnsw} (HNSW) Graph to perform a _k_ approximate nearest neighbors (_k_-ANN) query over the vector fields.
+
+A common semantic of a vector is as an embedding, a numerical representation of a resource, for example, text, images, audio, or documents.
```
JPryce-Aklundh:
```suggestion
A common semantic of a vector is as an embedding.
This is a numerical representation of a resource such as a text, image, audio, or document.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+Embedding generators such as https://cloud.google.com/vertex-ai[Vertex AI] or https://openai.com/[OpenAI] can produce 768, 1024, 1536, etc., dimensional vector embeddings.
```
JPryce-Aklundh: This is probably due to my ignorance, but do we need to elaborate on these numbers at all?

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+You can create vector indexes using the procedure {link-procedures-reference}#procedure_db_index_vector_createNodeIndex[`db.index.vector.createNodeIndex`].
+The name of the index must be unique so that you can reference it when xref:#indexes-vector-query[querying] or if you want to xref:indexes-vector-drop[drop] the index.
+
+[IMPORTANT]
+====
+The index name must be unique among both indexes and constraints.
+====
+
+A vector index is a single-label, single-property index for nodes.
+In addition to the label and property key, both given as `STRING`, a vector index needs to be configured with both the dimensionality of the vector, `INTEGER`, and the measure of similarity between two vectors, case-insensitive `STRING` For details, see xref:#indexes-vector-similarity[].
```
JPryce-Aklundh:
```suggestion
A vector index is a single-label, single-property index for nodes.
In addition to the label and property key (both given as `STRING`), a vector index needs to be configured with both the dimensionality of the vector (`INTEGER`), and the measure of similarity between two vectors (case-insensitive `STRING`).
For details, see xref:#indexes-vector-similarity[].
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+You can query a vector index using the procedure {link-procedures-reference}#procedure_db_index_vector_queryNodes[`db.index.vector.queryNodes`].
+
+.Signature for `db.index.vector.queryNodes` to query a vector index
+[source, syntax, role="noheader", indent=0]
+----
+db.index.vector.queryNodes(indexName :: STRING?, numberOfNearestNeighbours :: INTEGER?, query :: LIST? OF FLOAT?) :: (node :: NODE?, score :: FLOAT?)
+----
+
+* The `indexName` (a string) refers to the unique name of the vector index to query.
+* The `numberOfNearestNeighbours` (an integer) refers to the number of nearest neighbors to return as the neighborhood.
```
JPryce-Aklundh:
```suggestion
* The `numberOfNearestNeighbours` (an `INTEGER`) refers to the number of nearest neighbors to return as the neighborhood.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+* `LIST<FLOAT>` query vector in which to search for the neighborhood.
+
+The procedure returns the neighborhood of nodes with their respective similarity scores, ordered by those scores.
+The scores are bounded between `0` and `1`, where the closer to `1`, the more similar to the query vector.
```
JPryce-Aklundh:
```suggestion
The scores are bounded between `0` and `1`, where the closer to `1` the score is, the more similar to the query vector the neighborhood of nodes is.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+
+.Query vector index
```
JPryce-Aklundh:
```suggestion
.Query a vector index
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+====
+For instance, let's take the paper that describes the HNSW{fn-hnsw} graph structure that the vector index implements, and try and find similar papers.
```
JPryce-Aklundh:
```suggestion
This example takes the paper that describes the HNSW{fn-hnsw} graph structure that the vector index implements and tries to find similar papers.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+First you `MATCH` to find the paper, you then query the `abstract-embeddings` index for a neighborhood of `10` similar abstracts to your query.
```
JPryce-Aklundh:
```suggestion
First you `MATCH` to find the paper, and then you query the `abstract-embeddings` index for a neighborhood of `10` similar abstracts to your query.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+| 0.9267876148223877
+
+| "Learning to Route in Similarity Graphs"
+| 0.9263730049133301
+
+2+d| Rows: 10
+
+|===
+
+The results are quite expected, with papers discussing graph-based nearest-neighbor searches.
```
JPryce-Aklundh:
```suggestion
The results are expected, with papers discussing graph-based nearest-neighbor searches.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+
+The most similar to this result is the query vector itself, which is to be expected as you queried the index with an indexed property.
```
JPryce-Aklundh:
```suggestion
The most similar to this result is the query vector itself, which is to be expected as the index was queried with an indexed property.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+The {link-procedures-reference}#procedure_db_create_setVectorProperty[`db.create.setVectorProperty`] procedure is provided to validate and set the property as an array of {ieee-754} single precision values.
+As a result, the store's space is approximately halved.
+
+.Signature for `db.create.setVectorProperty` to set a vector property
+[source, syntax, role="noheader", indent=0]
+----
+db.create.setVectorProperty(node :: NODE?, key :: STRING?, vector :: LIST? OF FLOAT?) :: (node :: NODE?)
+----
+
+Thus, the equivalent syntax for setting as the aforementioned:
```
JPryce-Aklundh:
```suggestion
Thus, this is the equivalent syntax for setting a vector with a procedure:
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+CALL db.create.setVectorProperty(node, 'propertyKey', vector)
+----
+
+This is ideal for creating and setting new vector properties in the graph.
+Though existing properties can be re-set using this procedure to minimize store size, there is a cost in transaction log size until they are rotated away, especially in the case of record formats.
+
+[[indexes-vector-similarity]]
+== Supported similarity functions
+
+The choice of similarity function can vastly differ in which vectors are considered similar, and which are valid.
```
JPryce-Aklundh:
```suggestion
The choice of similarity function can differ vastly depending on which vectors are considered similar, and which are valid.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+The semantic meaning of the vector may itself dictate which similarity function to choose.
+Consult the documentation for the particular model used to generate vector embeddings.
+The documentation may suggest a preference for certain similarity functions.
```
JPryce-Aklundh:
```suggestion
Consult the documentation for the particular model used to generate vector embeddings, as this may suggest a preference for certain similarity functions.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+Otherwise, understanding the differences can help make an informed choice.
```
JPryce-Aklundh:
```suggestion
Otherwise, understanding the differences between different similarity functions can help make a more informed choice.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+| "euclidean"
+| distance
+
+| xref:indexes-vector-similarity-cosine[Cosine]
+| "cosine"
+| angle
+
+|===
+
+For {l2-norm}alized vectors, such that their length, image:l2norm_is_1.svg["The l2-norm of vector v equals 1"], both Euclidean and cosine similarity functions produce the same similarity ordering.
```
JPryce-Aklundh: Is this still accurate?
```suggestion
For {l2-norm}alized vectors, where the length is image:l2norm_is_1.svg["The l2-norm of vector v equals 1"], both Euclidean and cosine similarity functions produce the same similarity ordering.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+
+[[indexes-vector-similarity-euclidean]]
+=== Euclidean similarity
+
+Euclidean is useful when the _distance_ between the vectors is what determines how similar two vectors are.
```
JPryce-Aklundh:
```suggestion
Euclidean similarity is useful when the _distance_ between the vectors is what determines how similar two vectors are.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+* And its {l2-norm} is non-zero and can be represented finitely in {ieee-754} single precision.
+
+Cosine interprets the vectors in Cartesian coordinates.
+The measure is related to the angle between the two vectors.
+However, an angle can be described in many units, sign conventions, and periods.
+The trigonometric cosine of this angle is both agnostic to the aforementioned and bounded.
+Cosine similarity rebounds the trigonometric cosine.
+
+image::cosine_similarity_equation.svg["The cosine of vector v and vector u is defined as half of the quanity 1 plus the scalar product of v hat u hat, which equals half of the quantity 1 plus the scalar product of vector v vector u over the product of the l2-norm of vector v and the l2 norm ov vector u, which exists in the bounded set of real numbers between 0 inclusive and 1 inclusive."]
+Where the trigonometric cosine is given by the scalar product of two unit vectors.
```
JPryce-Aklundh: What is this "Where" referring to? Could it be replaced with something like: "The trigonometric cosine in the above equation..."?

```suggestion
Where the trigonometric cosine is given by the scalar product of two unit vectors.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+
+[[indexes-vector-limitations]]
+== Known limitations and quirks
```
JPryce-Aklundh:
```suggestion
== Limitations and idiosyncrasies
```
This could maybe just be "Limitations"?

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+
+:link-operations-manual: link:{neo4j-docs-base-uri}/operations-manual/{page-version}
+:link-procedures-reference: {link-operations-manual}/reference/procedures/
+
+:ieee-754: IEEE 754{fn-ieee-754}
+
+:l2-norm: image:l2.svg["l2"]-norm
+
+[[indexes-vector]]
+= Nearest neighbor search
```
JPryce-Aklundh: Could we call the chapter Vector indexes instead?

JPryce-Aklundh (commented):
Thanks @parnmatt -  I have made some editorial suggestions, and also have a few questions where I am unsure about the meaning of something.

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
renetapopova: Hey @parnmatt , which documentation do you refer to? Could you provide us with a link?

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+
+[abstract]
+--
+This chapter describes how to use vector indexes to perform an approximate nearest neighbor search.
+--
+
+_This feature is introduced in Neo4j 5.11._ label:beta[]
+
+Vector indexes are powered by the link:https://lucene.apache.org/[Apache Lucene] indexing and search library, and can be used to index nodes by `LIST<FLOAT>` properties valid to the index.
```
JPryce-Aklundh: I know the full-text index page currently starts like this (I have not yet had the time to fix it), but I think it is not the most helpful way to start. Maybe moving forward, we could add an introduction section (including answers to questions such as "what are vector indexes?", "why would you use them?") to start the page with, to ease the learning curve for those not already initiated.

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
stefano-ottolenghi: @parnmatt yes, see here https://github.com/neo4j/docs-testing/tree/main/test-cypher
TL;DR: don't mix `[source, cypher]` with `[source, syntax]`, as all cypher blocks get executed for testing.

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
```diff
@@ -0,0 +1,431 @@
...
+----
+SHOW INDEXES YIELD name, state, type, labelsOrTypes, properties, options
+WHERE type = "VECTOR"
+----
+
+.Result
+[role="queryresult",options="header,footer",cols="3m,2m,3m,3m,6a"]
+|===
+
+| name | type | labelsOrTypes | properties | options
```
stefano-ottolenghi: The `state` column is missing from the result, which causes a failing test.

stefano-ottolenghi (dismissed):

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: We could; I was just following from the full text index page, which was named for the query it enables, not the index type … though they are effectively the same, they are semantically different
full text index --> full text search
vector index --> (approximate) nearest neighbor search

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: Yeah I tooks some inspiration from full text as they're somewhat similar in their own ways being somewhat separate from the other indexes.

Sure, absolutely; as we've discussed in the past for a index/constraints rewrite, most of the actual _details_ of the indexes will likely be moved to operations manual. But those discussions are likely to be on a more general "index" page, rather than a specific index page … though I guess "why would you use _this_ index" might be relevant to bring forward.

Do we want to have such a section in this first iteration? Provide it later? Or when we eventually find time to work on the index/constraint rewrite?

I say this, as 5.11 went live on Aura _this morning_

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: Examples of common embedding, and thus vector sizes.
Just an implicit note that suggests to a less knowledgable user what "number" to use from the several that might be thrown at them from certain generators or their documentation.

It also implies a scale difference. It's highly likely we're not working with 2 or 3 dimensional vectors (like `POINT`), but hundreds/thousands of dimensions. A very different beast.

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: AH, sorry, that's my copy/paste issue
I originally had many more columns, but they weren't nicely rendering, so I reduced them :)

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: the score is individual to the indexed vector compared to the query vector
So it doesn't make sense to say "neighborhood" in this case, as that implies to me the score is over the whole neighborhood.

That would be a different kind of query, if I were to ask for all vectors that are within a score of 0.9 to 1.
Which is unsupported.

However I agree this could do with a similar reword.
How about:

```suggestion
The scores are bounded between `0` and `1`, where the closer to `1` the score is, the more similar the indexed vector is to the query vector.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: Ah, I see if the confusion in what I had originally written; the relationship should be reversed.

The vectors that are considered valid for indexing and querying with, and more importantly, the indexed vectors that are considered similar to a query vector, can differ vastly depending on the choice of similarity function.

@JPryce-Aklundh what would be your suggestion to reword the lines to better encode _that_?

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: Sure, will try to find the link

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: I am not sure about the use of "where" here. It _can_ make sense, but could also be read to indicate that there could be l2-normalised vectors where the length is not 1 … which is impossible.

What are your thoughts about:
```suggestion
For {l2-norm}alized vectors (unit vectors), thus their length image:l2norm_is_1.svg["The l2-norm of vector v equals 1"], both Euclidean and cosine similarity functions produce the same similarity ordering.
```

modules/ROOT/pages/indexes-for-nearest-neighbor.adoc
parnmatt: _where_ is often used in academic writing following an equation specifically to clarify parts of it; often in lowercase as it's considered a continuation of the equation's sentence.

There is no standard function notation for the angle between two vectors in terms of those two vectors; thus cannot write something like `cos(angleBetween(v,u))`, and have to write `v.u` which is more common.

Sure, I wouldn't be opposed to something like
```suggestion
The trigonometric in the above equation is given by the scalar product of the two unit vectors.
```
Or, more preferably,
```suggestion
where scalar product of the two unit vectors is the trigonometric cosine of the angle between the two vectors.
```
or some hybrid. What are your thoughts and suggestions? Whichever we go with, I will make the same change in the operations manual which also has these.

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -21,10 +21,11 @@ An embedding is a numerical representation of a data object, such as text, image
...
-These dimensional vector embeddings are stored as `LIST<FLOAT>` properties in a Neo4j vector index.
-(It can be used to index nodes by `LIST<FLOAT>` properties valid to the index.)
+This allows for mathematical operations like addition and subtraction to carry semantic meaning; for example, the vector representation of "king" minus "man" plus "woman" might be close to the vector representation of "queen." In other words, vector embeddings can be said to be a numerical representation of a particular data object, capturing its semantic meaning.
+
+The embedding for a particular data object can be generated by for example the https://cloud.google.com/vertex-ai[Vertex AI] or https://openai.com/[OpenAI] embedding generators, which can produce vector embeddings with dimensions 768, 1024, 1536, etc.
+These vector embeddings are stored as `LIST<FLOAT>` properties on a node, where each dimensional component of the vector, is an element in the `LIST`.
+A Neo4j vector index can be used to index nodes by `LIST<FLOAT>` properties valid to the index.
```
parnmatt, line 30: This last sentence may want a slight reword. It feels like there are too many `index`es

parnmatt:
@mauvo also requested we change the title; however, to "Vector search index" to align with marketing, and the "Full-text search index" page, etc.

Thus that was included, and the file renamed accordingly for consistency, and so that he could have the effective link for the release notes ready.

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,432 @@
...
+
+======
+
+[[indexes-vector-set]]
+== Set vector property on a node
+
+You can set a vector property, `LIST<FLOAT>`, on a node using the xref:clauses/set.adoc#set-set-a-property[`SET`] command:
+
+.Syntax for setting a vector property
+[source,cypher]
```
```suggestion
[source, syntax]
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,432 @@
...
+.Signature for `db.create.setVectorProperty` to set a vector property
+[source,syntax,role="noheader",indent=0]
+----
+db.create.setVectorProperty(node :: NODE?, key :: STRING?, vector :: LIST? OF FLOAT?) :: (node :: NODE?)
+----
+
+Thus, this is the equivalent syntax for setting a vector with a procedure:
+
+.Syntax for setting a vector _via_ a procedure
+[source,cypher]
```
```suggestion
[source, syntax]
```

JPryce-Aklundh (commented):
I believe fixing these two will fix the failing tests

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+An embedding is a numerical representation of a data object, such as text, image, audio, or document.
+
+For example, each word or token in a text is typically represented as high-dimensional vector where each dimension represents a certain aspect of the word’s meaning.
+Words that are semantically similar or related are often represented by vectors that are closer to each other in this vector space.
+This allows for mathematical operations like addition and subtraction to carry semantic meaning; for example, the vector representation of "king" minus "man" plus "woman" might be close to the vector representation of "queen." In other words, vector embeddings can be said to be a numerical representation of a particular data object, capturing its semantic meaning.
+
+The embedding for a particular data object can be generated by for example the https://cloud.google.com/vertex-ai[Vertex AI] or https://openai.com/[OpenAI] embedding generators, which can produce vector embeddings with dimensions 768, 1024, 1536, etc.
+These vector embeddings are stored as `LIST<FLOAT>` properties on a node, where each dimensional component of the vector, is an element in the `LIST`.
```
renetapopova:
```suggestion
These vector embeddings are stored as `LIST<FLOAT>` properties on a node, where each dimensional component of the vector is an element in the `LIST`.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+_This feature was introduced in Neo4j 5.11._ label:beta[]
+
+This chapter describes how to use vector indexes to perform an approximate nearest neighbor search.
+
+Vector indexes allow users to query vector embeddings from large datasets.
+An embedding is a numerical representation of a data object, such as text, image, audio, or document.
+
+For example, each word or token in a text is typically represented as high-dimensional vector where each dimension represents a certain aspect of the word’s meaning.
+Words that are semantically similar or related are often represented by vectors that are closer to each other in this vector space.
+This allows for mathematical operations like addition and subtraction to carry semantic meaning; for example, the vector representation of "king" minus "man" plus "woman" might be close to the vector representation of "queen." In other words, vector embeddings can be said to be a numerical representation of a particular data object, capturing its semantic meaning.
```
renetapopova:
```suggestion
This allows for mathematical operations like addition and subtraction to carry semantic meaning.
For example, the vector representation of "king" minus "man" plus "woman" might be close to the vector representation of "queen." 
In other words, vector embeddings can be said to be a numerical representation of a particular data object, capturing its semantic meaning.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+
+The embedding for a particular data object can be generated by for example the https://cloud.google.com/vertex-ai[Vertex AI] or https://openai.com/[OpenAI] embedding generators, which can produce vector embeddings with dimensions 768, 1024, 1536, etc.
```
renetapopova:
```suggestion
The embedding for a particular data object can be generated by for example the https://cloud.google.com/vertex-ai[Vertex AI] or https://openai.com/[OpenAI] embedding generators, which can produce vector embeddings with dimensions 768, 1024, and 1536.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+You want to find papers in the neighborhood of a paper you know.
+
+.Data model
+[source,syntax,role="noheader",indent=0]
+----
+(:Title)<--(:Paper)-->(:Abstract)
+----
+
+Assume for each abstract, you have generated a 1536-dimensional vector `embedding` of the abstract's `text` using Open AI's default model, `text-embedding-ada-002`.
+This model suggests a xref:indexes-vector-similarity-cosine[cosine similarity] (for more information see OpenAI's documentation on https://platform.openai.com/docs/guides/embeddings/which-distance-function-should-i-use[Which distance function I use?]).
```
renetapopova:
```suggestion
This model suggests a xref:indexes-vector-similarity-cosine[cosine similarity].
For more information, see link:https://platform.openai.com/docs/guides/embeddings/which-distance-function-should-i-use[OpenAI's official documentation].
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+CALL db.index.vector.createNodeIndex('abstract-embeddings', 'Abstract', 'embedding', 1536, 'cosine')
+----
+
+.Result
+[queryResult]
+----
+(no changes, no records)
+----
+
+You can see that the vector index has been created using `SHOW INDEXES`
```
renetapopova:
```suggestion
You can see that the vector index has been created using `SHOW INDEXES`:
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+.Result
+[queryresult]
+----
+Removed 1 index.
+----
+
+======
+
+[[indexes-vector-set]]
+== Set vector property on a node
```
renetapopova:
```suggestion
== Set a vector property on a node
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+
+This is ideal for creating and setting new vector properties in the graph.
+Though existing properties can be re-set using this procedure to minimize store size, there is a cost in transaction log size until they are rotated away, especially in the case of record formats.
+
+[[indexes-vector-similarity]]
+== Supported similarity functions
+
+The choice of similarity function can vastly differ in which indexed vectors are considered similar, and which are valid.
+The semantic meaning of the vector may itself dictate which similarity function to choose.
+Consult the documentation for the particular model used to generate vector embeddings.
```
renetapopova:
```suggestion
Refer to the documentation for the particular vector embedding model you are using, as it may suggest a preference for certain similarity functions.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+The documentation may suggest a preference for certain similarity functions.
```
renetapopova:
```suggestion
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+Otherwise, understanding the differences between different similarity functions can help make a more informed choice.
```
renetapopova:
```suggestion
Otherwise, being able to differentiate between the various similarity functions can assist in making a more informed decision.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+
+.Similarity functions
+[%header,cols="d,m,e"]
+|===
+| Name | Case Insensitive Argument | Key similarity feature
```
renetapopova:
```suggestion
| Name | Case insensitive argument | Key similarity feature
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+| "euclidean"
+| distance
+
+| xref:indexes-vector-similarity-cosine[Cosine]
+| "cosine"
+| angle
+
+|===
+
+For {l2-norm}alized vectors (unit vectors), thus their length image:l2norm_is_1.svg["The l2-norm of vector v equals 1"], both Euclidean and cosine similarity functions produce the same similarity ordering.
```
renetapopova:
```suggestion
For {l2-norm}alized vectors (unit vectors), thus their length image:l2norm_is_1.svg["The l2-norm of vector v equals 1"], Euclidean and cosine similarity functions produce the same similarity ordering.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+[[indexes-vector-similarity-euclidean]]
+=== Euclidean similarity
+
+Euclidean similarity is useful when the _distance_ between the vectors is what determines how similar two vectors are.
+
+A valid vector for a Euclidean vector index is when all vector components can be represented finitely in {ieee-754} single precision.
+
+Euclidean interprets the vectors in Cartesian coordinates.
+The measure is related to the Euclidean distance, how far two points are from one another.
+However, that is unbounded, and less useful as a similarity score.
```
renetapopova:
```suggestion
However, that is unbounded and less useful as a similarity score.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -0,0 +1,433 @@
...
+Euclidean similarity bounds the square of Euclidean distance.
+
+image::euclidean_similarity_equation.svg["The Euclidean of vector v and vector u is defined as 1 over the quantity 1 plus the square of the l2-norm of vector v subtract vector u, which exists in the bounded set of real numbers between 0 exclusive and 1 inclusive."]
+
+[[indexes-vector-similarity-cosine]]
+=== Cosine similarity
+
+Cosine should be used when the _angle_ between the vectors is what determines how similar two vectors are.
```
renetapopova:
```suggestion
Cosine is used when the _angle_ between the vectors is what determines how similar two vectors are.
```

renetapopova (commented):
Thanks, @parnmatt. I added some comments.

JPryce-Aklundh (approved):

parnmatt:
> Thanks, @parnmatt. I added some comments.

Thanks, I was about to have a look, but it seems @JPryce-Aklundh has beaten me to it, and did them individually rather than a batch 😅 

In that case, I'll have a skim of the changes, and they can always be changed later if need be.

modules/ROOT/pages/indexes-for-vector-search.adoc
parnmatt: That's now a finite list; different models they provide can produce even more than what was listed (hence the _etc_), but those are the common ones from those generators.

renetapopova: Yes, that's why I changed it to a finite list, because here we give those two as examples.

parnmatt: I don't know if I was clear, sure there are several models, and several venders, and I listed 2 venders, listing the 3 most common vector dimensions that they'd offer; but they offer _more_ than just those 3.

For completeness
I think VertexAI can do 768 for text and 1408 for multimodal, they used to have 1024 for images, but I cannot find that in their docs anymore.
OpenAI gen1 models could be 1024, 2048, 4096, 12288 ... and their now prefered/recommended gen2 is 1536

so I don't know if a finite list is the right call.
perhaps with a "e.g." (or I guess "for example") as that implies it's not a complete list without using "etc."

PR #700, opened 2023-08-21
Update vector index issues

parnmatt:


neo-technology-commit-status-publisher:

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -387,37 +387,78 @@ For example, you cannot have one xref:indexes-vector-similarity-euclidean[Euclid
...
+| The creation of the vector index skipped the check to limit the dimensionality to `2048`.
 
-| The validation for xref:indexes-vector-similarity-cosine[cosine similarity] verifies the vector's {l2-norm} can be represented finitely in {ieee-754} *double* precision, rather than _single_ precision.
+[NOTE]
+====
+Vector indexes configured with dimensionality greater than `2048` in Neo4j 5.11 should continue to work after the limitation is applied.
+====
+|
+
+d| The validation for xref:indexes-vector-similarity-cosine[cosine similarity] verifies the vector's {l2-norm} can be represented finitely in {ieee-754} *double* precision, rather than _single_ precision.
```
```suggestion
d| The validation for xref:indexes-vector-similarity-cosine[cosine similarity] verifies that the vector's {l2-norm} can be represented finitely in {ieee-754} *double* precision, rather than in _single_ precision.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -387,37 +387,78 @@ For example, you cannot have one xref:indexes-vector-similarity-euclidean[Euclid
...
 This can lead to certain large component vectors being incorrectly indexed, and return a similarity score of `&plusmn;0.0`.
 a| // Neo4j 5.12    pending PR
 
-| The vector index `createStatement` field from xref:indexes-for-search-performance.adoc#indexes-list-indexes[`SHOW INDEXES`] does not correctly escape single quotes in index names, labels, and property keys.
+| {link-procedures-reference}#procedure_db_index_vector_queryNodes[`db.index.vector.queryNodes`] query vector validation is incorrect with a xref:indexes-vector-similarity-cosine[cosine] vector index. The {l2-norm} validation only considers the last component of the vector. If that component is `&plusmn;0.0`, an otherwise valid query vector will be thrown as invalid. This can also result in some invalid vectors being used to query, and return a similarity scores of `&plusmn;0.0`.
+
+[TIP]
+====
+For {l2-norm}alized vectors (unit vectors), thus their length image:l2norm_is_1.svg["The l2-norm of vector v equals 1"], Euclidean and cosine similarity functions produce the same similarity ordering.
```
JPryce-Aklundh: The "..., thus their length...," part reads awkwardly to me. Could we bracket it, or rephrase it to something like "with length ..."? I think I made a similar suggestion in the first PR, but can't remember what the outcome was. It may be that this is the proper formating, and that I have not come across it before (and need to get used to it).

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -387,37 +387,78 @@ For example, you cannot have one xref:indexes-vector-similarity-euclidean[Euclid
...
 ERROR: [StoreCopy] Unable to format statement for index 'index-name'
 ----
 Due to an:
 ----
 java.lang.IllegalArgumentException: Did not recognize index type VECTOR
 ----
+
+[TIP]
+====
+If a store copy is required, please make note of the creation commands from:
```
```suggestion
If a store copy is required, make a note of the information in the `createStatement` column returned from the `SHOW INDEX` command.
For example:
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -387,37 +387,78 @@ For example, you cannot have one xref:indexes-vector-similarity-euclidean[Euclid
...
+[source,cypher]
+----
+SHOW INDEXES YIELD type, createStatement
+WHERE type = 'VECTOR'
+RETURN createStatement
+----
+====
+| Neo4j 5.12
+
+| Some of the protections preventing using new features during database rolling upgrade are missing. This can result in a transaction to create a vector index being created on a cluster member running a Neo4j 5.11 version and being distributed to other cluster members on an older Neo4j version. The older Neo4j versions will fail to understand the transaction, and will panic.
```
JPryce-Aklundh: Is "will panic" a formal way of describing a particular behaviour? We can keep it in if it is.
```suggestion
| Some of the protections preventing the use of new features during a database rolling upgrade are missing.
This can result in a transaction to create a vector index on a cluster member running Neo4j 5.11 and distributing it to other cluster members running an older Neo4j version.
The older Neo4j versions will fail to understand the transaction.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -387,37 +387,78 @@ For example, you cannot have one xref:indexes-vector-similarity-euclidean[Euclid
...
 | Passing `null` as an argument to some of the procedure parameters can generate a confusing exception.
-a| Neo4j 5.12
+| Neo4j 5.12
+
+| The creation of the vector index skipped the check to limit the dimensionality to `2048`.
 
-| The validation for xref:indexes-vector-similarity-cosine[cosine similarity] verifies the vector's {l2-norm} can be represented finitely in {ieee-754} *double* precision, rather than _single_ precision.
+[NOTE]
+====
+Vector indexes configured with dimensionality greater than `2048` in Neo4j 5.11 should continue to work after the limitation is applied.
```
JPryce-Aklundh, line 415: "should" sounds a bit uncertain, but maybe that is intentional?
```suggestion
Vector indexes configured with dimensionality greater than `2048` in Neo4j 5.11 will continue to work after the limitation is applied.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -387,37 +387,78 @@ For example, you cannot have one xref:indexes-vector-similarity-euclidean[Euclid
...
+====
+|
+
+d| The validation for xref:indexes-vector-similarity-cosine[cosine similarity] verifies the vector's {l2-norm} can be represented finitely in {ieee-754} *double* precision, rather than _single_ precision.
 This can lead to certain large component vectors being incorrectly indexed, and return a similarity score of `&plusmn;0.0`.
 a| // Neo4j 5.12    pending PR
 
-| The vector index `createStatement` field from xref:indexes-for-search-performance.adoc#indexes-list-indexes[`SHOW INDEXES`] does not correctly escape single quotes in index names, labels, and property keys.
+| {link-procedures-reference}#procedure_db_index_vector_queryNodes[`db.index.vector.queryNodes`] query vector validation is incorrect with a xref:indexes-vector-similarity-cosine[cosine] vector index. The {l2-norm} validation only considers the last component of the vector. If that component is `&plusmn;0.0`, an otherwise valid query vector will be thrown as invalid. This can also result in some invalid vectors being used to query, and return a similarity scores of `&plusmn;0.0`.
```
```suggestion
| {link-procedures-reference}#procedure_db_index_vector_queryNodes[`db.index.vector.queryNodes`] query vector validation is incorrect with a xref:indexes-vector-similarity-cosine[cosine] vector index. The {l2-norm} validation only considers the last component of the vector. If that component is `&plusmn;0.0`, an otherwise valid query vector will be thrown as invalid. This can also result in some invalid vectors being used to query, and return a similarity score of `&plusmn;0.0`.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -387,37 +387,78 @@ For example, you cannot have one xref:indexes-vector-similarity-euclidean[Euclid
...
+RETURN createStatement
+----
+====
+| Neo4j 5.12
+
+| Some of the protections preventing using new features during database rolling upgrade are missing. This can result in a transaction to create a vector index being created on a cluster member running a Neo4j 5.11 version and being distributed to other cluster members on an older Neo4j version. The older Neo4j versions will fail to understand the transaction, and will panic.
+
+[TIP]
+====
+Ensure that all cluster members have been updated to use Neo4j 5.11 or newer before calling `dbms.upgrade()` on the `system` database. Once commited, vector indexes can be safely created on the cluster.
```
```suggestion
Ensure that all cluster members have been updated to use Neo4j 5.11 (or a newer version) before calling `dbms.upgrade()` on the `system` database. Once committed, vector indexes can be safely created on the cluster.
```

modules/ROOT/pages/indexes-for-vector-search.adoc
nilsceberg: **double** is bold and *single* is italic, is that intended?

JPryce-Aklundh (commented):
Thanks @parnmatt - I have made a few small editorial suggestions, and also have a few questions. Also, if possible, could you start each new sentence on a new line in the asciidoc file?

modules/ROOT/pages/indexes-for-vector-search.adoc
nilsceberg: Maybe "preventing *the use of* new features"

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -387,37 +387,78 @@ For example, you cannot have one xref:indexes-vector-similarity-euclidean[Euclid
...
+====
+The procedures are still usable, just not visible.
+====
 | Neo4j 5.12
 
 | Passing `null` as an argument to some of the procedure parameters can generate a confusing exception.
-a| Neo4j 5.12
+| Neo4j 5.12
+
+| The creation of the vector index skipped the check to limit the dimensionality to `2048`.
```
nilsceberg, line 411: I guess we could also document 2048 as being the limit for what is *supported* and that it *might* be a hard limit in the future... this list is getting long 😄

nilsceberg (commented):
Looks geat to me :+1:

modules/ROOT/pages/indexes-for-vector-search.adoc
parnmatt: I believe this was in the original.
It was, different levels of emphasis.

modules/ROOT/pages/indexes-for-vector-search.adoc
parnmatt: I chose 'should' as an indication that it is not supported.
We _may_ add more rigorous correctness checks in the future that does prevent these older beta indexes from working.
So I'm not sure it should be `will`, though I am still open to change it

modules/ROOT/pages/indexes-for-vector-search.adoc
parnmatt: this was the compromise from before (as I copy/paste it).
Normally it would be common to have a comma between a concept and their mathematical notation `the radius of a circle, r, is the length…`, and if there are multiple listed things, you use comma and semi-colons (as a super-comma).
But that is common in academic writing … which this is not (really).

You guys had issue with the original comma and removed it.
I think perhaps we could go with:

```suggestion
For {l2-norm}alized vectors (unit vectors), thus having unit length image:l2norm_is_1.svg["The l2-norm of vector v equals 1"], Euclidean and cosine similarity functions produce the same similarity ordering.
```

I'd still want the comma, and perhaps a semi-colon; however, I know you guys don't like them :sweat_smile:

modules/ROOT/pages/indexes-for-vector-search.adoc
parnmatt: yes, `panic` is the correct term.

modules/ROOT/pages/indexes-for-vector-search.adoc
parnmatt: I'll think about it

modules/ROOT/pages/indexes-for-vector-search.adoc
JPryce-Aklundh: Your suggestion works 👍

modules/ROOT/pages/indexes-for-vector-search.adoc
nilsceberg: It was, just noticed it since the line had changed

modules/ROOT/pages/indexes-for-vector-search.adoc
nilsceberg: I'm not a big fan of letting old indexes continue working but preventing the creation of new ones because it turns indexes into data in a sense... as in you can't just wipe them and recreate them. That's probably a discussion for later, but I think committing to it here is a bad idea and if anything I think even "should" might promise too much. :)

JPryce-Aklundh: Ok - let's stick with "should" for now 👍

parnmatt:
@JPryce-Aklundh I believe we're a-go if you're happy

JPryce-Aklundh (approved):
lgtm 👍

PR #723, opened 2023-09-14
Improve Path Pattern Expression documentation

AzuObs:
https://trello.com/c/c0SxcchB/2009-clarify-restriction-of-single-node-pattern-in-where-clauses

### DIFF
--- a/modules/ROOT/pages/clauses/where.adoc
+++ b/modules/ROOT/pages/clauses/where.adoc
@@ -491,7 +491,7 @@ While existential subqueries are more powerful and capable of performing anythin
 Path pattern expressions have the following restrictions (use cases that require extended functionality should consider using xref::subqueries/existential.adoc[existential subqueries] instead):
 
 * Path pattern expressions may only use a subset of xref::patterns/reference.adoc#graph-patterns[graph pattern] semantics.
-A path pattern expressions must be a xref::patterns/reference.adoc#path-patterns[path pattern] of length greater than zero.
+* A path pattern expression must be a xref::patterns/reference.adoc#path-patterns[path pattern] of length greater than zero.
 In other words, it must contain at least one xref::patterns/reference.adoc#relationship-patterns[relationship] or xref::patterns/reference.adoc#variable-length-relationships[variable-length relationship].
 
 Path pattern expressions may not declare new variables.
--- a/modules/ROOT/pages/clauses/where.adoc
+++ b/modules/ROOT/pages/clauses/where.adoc
@@ -485,8 +485,8 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND' 
 [[path-pattern-expressions]]
 === About path pattern expressions
 
-Path pattern expressions can be used to assert whether a specified path exists at least once in a graph. Path pattern expressions are similar to xref::subqueries/existential.adoc[existential subqueries].
-On one hand, xref::subqueries/existential.adoc[existential subqueries] are the more powerful feature and can do anything path pattern expressions can, while on the other hand path pattern expressions are more concise.
+Similar to xref::subqueries/existential.adoc[existential subqueries], path pattern expressions can be used to assert whether a specified path exists at least once in a graph.
+While existential subqueries are more powerful and capable of performing anything achievable with path pattern expressions, path pattern expressions are more concise.
 
 Path pattern expressions come with several restrictions which are highlighted below.
 Use cases which require more permissive functionality should consider using xref::subqueries/existential.adoc[existential subqueries] instead.
--- a/modules/ROOT/pages/clauses/where.adoc
+++ b/modules/ROOT/pages/clauses/where.adoc
@@ -479,10 +479,8 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND' 
 2+|Rows: 1
 |===
 
-[[query-where-patterns]]
-== Path pattern expressions
-
 [[path-pattern-expressions]]
+== Path pattern expressions
 
 Similar to xref::subqueries/existential.adoc[existential subqueries], path pattern expressions can be used to assert whether a specified path exists at least once in a graph.
 While existential subqueries are more powerful and capable of performing anything achievable with path pattern expressions, path pattern expressions are more concise.
--- a/modules/ROOT/pages/clauses/where.adoc
+++ b/modules/ROOT/pages/clauses/where.adoc
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND' 
 [[filter-on-patterns]]
 === Filter on patterns
 
-Patterns are expressions in Cypher, expressions that return a list of paths.
-List expressions are also predicates; an empty list represents `false`, and a non-empty list represents `true`.
+Path patterns may be used both in a MATCH clause and in a WHERE clause. When used in a WHERE clause they are called path
+pattern expressions.
 
-Patterns are not only expressions, they are also predicates.
-The only limitation to a pattern is that it must be able to express it in a single path.
-Unlike `MATCH`, it is not possible to use commas between multiple paths.
-To combine multiple patterns using `WHERE`, instead use `AND`.
-Note that new variables cannot be introduced.
-Although it might look very similar to the `MATCH` patterns, the `WHERE` clause is used to eliminate matched paths.
-`MATCH (a)-[*]->(b)` is very different from `WHERE (a)-[*]->(b)`.
-The first will produce a path for every path it can find between `a` and `b`, whereas the latter will eliminate any matched paths where `a` and `b` do not have a directed relationship chain between them.
+Path patterns used in a MATCH clause should generally be thought of as retrieving data. Path patterns expressions used
+in a WHERE clause should on the other hand be thought of as eliminating data. The clause `MATCH (a)-[]->(b)` is very
+different from `WHERE (a)-[]->(b)`. The first will produce a record for every path it can find between `a` and `b`,
+whereas the latter will eliminate records where `a` and `b` do not have a directed relationship between them.
 
+There are several limitations for what path patterns can do in contrast to their path pattern counterparts.
+
+- Path patterns in a MATCH clause can introduce new variables. Path pattern expressions in a WHERE clause cannot.
+
+- Path patterns in a MATCH clause can enjoy the full range of xref::patterns/reference.adoc[Pattern] semantics. Path
+pattern expressions in a WHERE clause on the other hand can only use
+xref::patterns/reference.adoc#relationship-patterns[Relationships] and
+xref::patterns/reference.adoc#variable-length-relationships[Variable-length relationships] semantics.
+
+- Multiple path patterns in a MATCH clause can be comma separated. Multiple path pattern expressions in a WHERE clause
+can only be separated by xref::syntax/operators.adoc#query-operators-boolean[Boolean operators] such as `AND` and `OR`.
 
 .Query
 [source, cypher]
--- a/modules/ROOT/pages/clauses/where.adoc
+++ b/modules/ROOT/pages/clauses/where.adoc
@@ -486,8 +486,8 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND' 
 [[filter-on-patterns]]
 === Filter on patterns
 
-Path patterns may be used both in a MATCH clause and in a WHERE clause. When used in a WHERE clause they are called path
-pattern expressions.
+xref:patterns/concepts.adoc#path-patterns[Path patterns] may be used both in a `MATCH` clause and in a `WHERE` clause.
+When used in a `WHERE` clause they are called path pattern expressions.
 
 Path patterns used in a MATCH clause should generally be thought of as retrieving data. Path patterns expressions used
 in a WHERE clause should on the other hand be thought of as eliminating data. The clause `MATCH (a)-[]->(b)` is very
--- a/modules/ROOT/pages/clauses/where.adoc
+++ b/modules/ROOT/pages/clauses/where.adoc
@@ -479,36 +479,29 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND' 
 2+|Rows: 1
 |===
 
-
 [[query-where-patterns]]
-== Using path patterns in `WHERE`
+== Path pattern expressions
+
+[[path-pattern-expressions]]
+=== About path pattern expressions
+
+Path pattern expressions can be used to assert whether a specified path exists at least once in a graph. Path pattern expressions are similar to xref::subqueries/existential.adoc[existential subqueries].
+On one hand, xref::subqueries/existential.adoc[existential subqueries] are the more powerful feature and can do anything path pattern expressions can, while on the other hand path pattern expressions are more concise.
+
+Path pattern expressions come with several restrictions which are highlighted below.
+Use cases which require more permissive functionality should consider using xref::subqueries/existential.adoc[existential subqueries] instead.
+
+Path pattern expressions may only use a subset of xref::patterns/reference.adoc#graph-patterns[graph patterns] semantics.
+A path pattern expressions must be a xref::patterns/reference.adoc#path-patterns[path pattern] of length greater than zero.
+In other words, it must contain at least one xref::patterns/reference.adoc#relationship-patterns[relationship] or xref::patterns/reference.adoc#variable-length-relationships[variable-length relationship].
+
+Path pattern expressions may not declare new variables.
+They can only reference existing variables.
+
+Path pattern expressions may only be used in positions where a xref:queries/expressions.adoc#boolean[boolean expression] is expected.
+The following sections will demonstrate how to use path pattern expressions in a `WHERE` clause.
 
 [[filter-on-patterns]]
-=== Filter on patterns
-
-xref:patterns/concepts.adoc#path-patterns[Path patterns] may be used both in a `MATCH` clause and in a `WHERE` clause.
-When used in a `WHERE` clause they are called path pattern expressions.
-
-Path patterns used in a `MATCH` clause should be thought of as retrieving data.
-Path pattern expressions used in a `WHERE` clause should be thought of as filtering data.
-
-For example, the clause `MATCH (a)-[]->(b)` is very different from `WHERE (a)-[]->(b)`.
-The first will produce a record for every path it can find between `a` and `b`.
-The latter will filter records where `a` and `b` do not have a directed relationship between them.
-
-There are several limitations to what path patterns used in a `MATCH` clause can do in contrast to path pattern expressions used in a `WHERE` clause.
-
-- Path patterns in a `MATCH` clause can introduce new variables.
-The same is not true for path pattern expressions in a `WHERE` clause.
-
-- All  xref::patterns/reference.adoc[pattern] semantics apply to path patterns in a `MATCH` clause.
-Path pattern expressions in a `WHERE` clause can only use
-xref::patterns/reference.adoc#relationship-patterns[relationship] and
-xref::patterns/reference.adoc#variable-length-relationships[variable-length relationship] semantics.
-
-- Multiple path patterns in a `MATCH` clause can be separated by commas.
-Multiple path pattern expressions in a `WHERE` clause can only be separated by xref::syntax/operators.adoc#query-operators-boolean[Boolean operators] such as `AND` and `OR`.
-
 === Filter on patterns
 
 .Query
@@ -517,7 +510,7 @@ Multiple path pattern expressions in a `WHERE` clause can only be separated by x
 MATCH
   (timothy:Person {name: 'Timothy'}),
   (other:Person)
-WHERE other.name IN ['Andy', 'Peter'] AND (other)-->(timothy)
+WHERE (other)-->(timothy)
 RETURN other.name, other.age
 ----
 
@@ -541,10 +534,10 @@ The `NOT` operator can be used to exclude a pattern:
 [source, cypher]
 ----
 MATCH
-  (person:Person),
-  (peter:Person {name: 'Peter'})
-WHERE NOT (person)-->(peter)
-RETURN person.name, person.age
+  (peter:Person {name: 'Peter'}),
+  (other:Person)
+WHERE NOT (other)-->(peter)
+RETURN other.name, other.age
 ----
 
 The `name` and `age` values for nodes that do not have an outgoing relationship to `Peter` are returned:
@@ -552,7 +545,7 @@ The `name` and `age` values for nodes that do not have an outgoing relationship 
 .Result
 [role="queryresult",options="header,footer",cols="2*<m"]
 |===
-| person.name | person.age
+| other.name | other.age
 | "Timothy" | 25
 | "Peter" | 35
 2+|Rows: 2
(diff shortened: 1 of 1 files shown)


AzuObs:
https://trello.com/c/c0SxcchB/2009-clarify-restriction-of-single-node-pattern-in-where-clauses

neo-technology-commit-status-publisher:

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
-Patterns are not only expressions, they are also predicates.
-The only limitation to a pattern is that it must be able to express it in a single path.
-Unlike `MATCH`, it is not possible to use commas between multiple paths.
-To combine multiple patterns using `WHERE`, instead use `AND`.
-Note that new variables cannot be introduced.
-Although it might look very similar to the `MATCH` patterns, the `WHERE` clause is used to eliminate matched paths.
-`MATCH (a)-[*]->(b)` is very different from `WHERE (a)-[*]->(b)`.
-The first will produce a path for every path it can find between `a` and `b`, whereas the latter will eliminate any matched paths where `a` and `b` do not have a directed relationship chain between them.
+Path patterns may be used both in a MATCH clause and in a WHERE clause. When used in a WHERE clause they are called path
+pattern expressions.
```
JPryce-Aklundh:
```suggestion
xref:patterns/concepts.adoc#path-patterns[Path patterns] may be used both in a `MATCH` clause and in a `WHERE` clause.
When used in a `WHERE` clause they are called path pattern expressions.
```

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
 
+Path patterns used in a MATCH clause should generally be thought of as retrieving data. Path patterns expressions used
+in a WHERE clause should on the other hand be thought of as eliminating data. The clause `MATCH (a)-[]->(b)` is very
+different from `WHERE (a)-[]->(b)`. The first will produce a record for every path it can find between `a` and `b`,
+whereas the latter will eliminate records where `a` and `b` do not have a directed relationship between them.
```
JPryce-Aklundh:
```suggestion
Path patterns used in a `MATCH` clause should be thought of as retrieving data.
Path pattern expressions used in a `WHERE` clause should be thought of as eliminating data.

For example, the clause `MATCH (a)-[]->(b)` is very different from `WHERE (a)-[]->(b)`.
The first will produce a record for every path it can find between `a` and `b`.
The latter will eliminate records where `a` and `b` do not have a directed relationship between them.
```

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
+
+There are several limitations for what path patterns can do in contrast to their path pattern counterparts.
```
JPryce-Aklundh:
```suggestion
There are several limitations to what path patterns used in a `MATCH` clause can do in contrast to path pattern expressions used in a `WHERE` clause.
```

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
+
+- Path patterns in a MATCH clause can introduce new variables. Path pattern expressions in a WHERE clause cannot.
```
JPryce-Aklundh:
```suggestion
- Path patterns in a `MATCH` clause can introduce new variables.
The same is not true for path pattern expressions in a `WHERE` clause.
```

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
+
+- Path patterns in a MATCH clause can enjoy the full range of xref::patterns/reference.adoc[Pattern] semantics. Path
+pattern expressions in a WHERE clause on the other hand can only use
+xref::patterns/reference.adoc#relationship-patterns[Relationships] and
+xref::patterns/reference.adoc#variable-length-relationships[Variable-length relationships] semantics.
```
JPryce-Aklundh:
```suggestion
- All  xref::patterns/reference.adoc[pattern] semantics apply to path patterns in a `MATCH` clause.
Path pattern expressions in a `WHERE` clause can only use
xref::patterns/reference.adoc#relationship-patterns[relationship] and
xref::patterns/reference.adoc#variable-length-relationships[variable-length relationship] semantics.
```

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
+
+- Multiple path patterns in a MATCH clause can be comma separated. Multiple path pattern expressions in a WHERE clause
+can only be separated by xref::syntax/operators.adoc#query-operators-boolean[Boolean operators] such as `AND` and `OR`.
```
JPryce-Aklundh:
```suggestion
- Multiple path patterns in a `MATCH` clause can be separated by commas.
Multiple path pattern expressions in a `WHERE` clause can only be separated by xref::syntax/operators.adoc#query-operators-boolean[Boolean operators] such as `AND` and `OR`.

=== Filter on patterns
```

modules/ROOT/pages/clauses/where.adoc
JPryce-Aklundh: I'm thinking the below information might be more efficiently communicated in a two column table (path patterns/ path pattern expressions) - what do you think?

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
 [[filter-on-patterns]]
 === Filter on patterns
```
JPryce-Aklundh: I think this needs a more distinct header (I have added the "Filter on patterns" as a suggestion below (line 509)

```suggestion
=== Note on path patterns and path pattern expressions
```

JPryce-Aklundh (commented):
Thanks @AzuObs - I think this will be very useful to our readers. I have made a few editorial suggestions - let me know what you think 👍

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
-
-Patterns are not only expressions, they are also predicates.
-The only limitation to a pattern is that it must be able to express it in a single path.
-Unlike `MATCH`, it is not possible to use commas between multiple paths.
-To combine multiple patterns using `WHERE`, instead use `AND`.
-Note that new variables cannot be introduced.
-Although it might look very similar to the `MATCH` patterns, the `WHERE` clause is used to eliminate matched paths.
-`MATCH (a)-[*]->(b)` is very different from `WHERE (a)-[*]->(b)`.
-The first will produce a path for every path it can find between `a` and `b`, whereas the latter will eliminate any matched paths where `a` and `b` do not have a directed relationship chain between them.
+Path patterns may be used both in a MATCH clause and in a WHERE clause. When used in a WHERE clause they are called path
```
arnefischereit: There are more positions for path patterns than just the `WHERE` clause. E.g.:
- `exists(<path pattern>)`
- `CASE WHEN <path pattern> THEN ... END`

They in turn can be used in all places that accept expressions.

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -486,18 +486,25 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
+pattern expressions.
 
+Path patterns used in a MATCH clause should generally be thought of as retrieving data. Path patterns expressions used
+in a WHERE clause should on the other hand be thought of as eliminating data. The clause `MATCH (a)-[]->(b)` is very
```
arnefischereit: I find "eliminating" sounds a lot like we are `DELETE`ing. The purpose of `WHERE` in general is to _filter_  the results/data that we found with `MATCH`.

What do you think about contrasting retrieving (or maybe "enumerating") in `MATCH` with "testing"  in `WHERE`? One could probably also use/promote the `EXISTS { ... }` or `exists(...)` expression here to explain what the pattern does in `WHERE`.

modules/ROOT/pages/clauses/where.adoc
arnefischereit: I think you have a point @JPryce-Aklundh, but I understood the two exactly opposite:
```suggestion
There are several limitations to what path patterns expressions used in a `WHERE` clause can do in contrast to path pattern used in a `MATCH` clause.
```

arnefischereit (dismissed):
I think it important that this documentation is reworked. The information currently available

>Patterns are expressions in Cypher, expressions that return a list of paths.

is just incorrect.

I also like that you emphasize that pattern expressions are in fact path pattern expressions, aligning this documentation with the pattern chapter that was created for QPPs.

I think, stating that path pattern expressions can only being used in `WHERE` is too restrictive. And I am not a big fan of writing "the path pattern expression was eliminating data".

I like the idea about creating a table on path pattern expressions vs. path patterns in `MATCH` (as this is how we should probably phrase it).

Also:
I did not find that we disallow writing `MATCH (n) WHERE (n)`, which was part of the task of this card.

AzuObs:
For Arne's comment that I do not explicitly say we do not allow `MATCH (n) WHERE (n)`... This is true, but I do say the following which should meet requirements of the card. I think it is better to tell users the things which are allowed rather than list of the potential things which are not allowed, but if you still think it is important to clarify we don't allow node pattern then I will add it.

```
Path pattern expressions in a `WHERE` clause can only use
xref::patterns/reference.adoc#relationship-patterns[relationship] and
xref::patterns/reference.adoc#variable-length-relationships[variable-length relationship] semantics.
```

modules/ROOT/pages/clauses/where.adoc
AzuObs: I agree that "eliminating" was a poor choice of words. I think that "retrieving" and "filtering" are good though. The page already uses "filtering" in many places which means the user should have a grasp of the concept by the time he reaches this section.

AzuObs: Both Jens and you have offered some suggestions about linking to other parts of the documentation which have their own sections in the docs. It would feel a bit spaghetti to me to include something about those other sections in here, so I may extract the "path pattern expressions" to their section of the docs and then link to them from relevant places such as WHERE (this page), CASE WHEN (other page), exists function (other page). I would not mention the EXISTS subquery as that has no special handling of path pattern expressions - if the EXISTS subquery has a WHERE clause then it can accept a path pattern expression, but so can anything else with a WHERE clause.

AzuObs: I feel that using `EXISTS { p }` or `exists( p )` in here will be quite confusing for users. This is because `EXISTS { }` accepts a path pattern whereas `exists( )` accepts has a parameter which is not a boolean but works with a pattern expression anyways (this is the only exception to the rule about pattern expressions). 

Given that ultimately we want to get rid of `exists( p )` altogether, I was hoping to not have to dive into the exception and KISS instead. I find the exception a bit superfluous anyways (YMMV) because `exists( pattern-expression )` is semantically the same as simply `pattern-expression`. They are both an existence check. 

As far as I can see, the purpose of `exists( pattern-expression )` is either as syntactic sugar or to allow using a pattern expression in places where any expression can be used.

arnefischereit: I like "filtering". That goes well together with the rest of the text.
What I meant with the `EXISTS { ... }` expression is that `WHERE (a)-[]→(b)` can be thought of as a shorthand of `WHERE EXISTS { (a)-[]→(b) }`.

I guess, pattern expressions are independent enough from `WHERE` clauses that they may reside in a page of their own. But @JPryce-Aklundh probably has an opinion on extracting that section.

arnefischereit:
I am not sure I understand what "relationship semantics" are. To me it suggests that a pattern expression consists of one relationship but there may be several.

I understand that having a positive definition instead of a negative one is preferable. You could for example refer to [path patterns](https://neo4j-docs-cypher-723.surge.sh/cypher-manual/5/patterns/reference/#path-patterns) and say that the path pattern should at least have one relationship.

For QPPs we fare this way:
> The path pattern being quantified must have a length greater than zero. In other words, it must contain at least one relationship. A single node pattern cannot be quantified.

modules/ROOT/pages/clauses/where.adoc
arnefischereit: You are right, `exists ( pattern-expression )` is nothing we want to promote. We mostly keep it as a legacy and to not unnecessarily introduce breaking changes.

I am not sure I understand your point about `EXISTS { ... }`, though: `EXISTS { ... }` is the more powerful construct and can accept arbitrary queries, including what can be expressed as path pattern expression. My idea was to frame path pattern expression as a short-hand to this. That way we promote `EXISTS { ... }` as a means when people want to express more than what can be expressed in a path pattern expression. And if people know `EXISTS { ... }` already, then such an equivalence will make it easier for them to learn path pattern expressions.

arnefischereit: This is just an idea, of course, so feel free to ignore it.

AzuObs: > I guess, pattern expressions are independent enough from WHERE clauses that they may reside in a page of their own. But @JPryce-Aklundh probably has an opinion on extracting that section.

I am going to try to make it work by keeping it on the current page. Other options I considered was moving it to the ["Cypher Expressions"](https://neo4j.com/docs/cypher-manual/current/queries/expressions/) page, or somewhere in the ["Cypher Patterns"](https://neo4j.com/docs/cypher-manual/current/patterns/) pages. All the options have pros and cons. I am finding it hard to get a clean result, but hopefully we can reach something that's good enough - certainly something that is better than what we had before.

modules/ROOT/pages/clauses/where.adoc
AzuObs: I went back and forth a bit, and settled on "About path pattern expressions"

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -555,49 +560,21 @@ Properties can also be added to patterns:
...
-.Result
-[role="queryresult",options="header,footer",cols="2*<m"]
-|===
-| type(r) | r.since
-| "KNOWS" | 1999
-| "KNOWS" | 2012
-2+|Rows: 2
-|===
-
-
```
AzuObs, line 600: This did not belong in this section, as it has nothing to do with path pattern expressions. 

If we need a new section to discuss the `type` function and its siblings `types`, `label`, and `labels`, then I would suggest making a new card for them. There however is already documentation concerning these on the https://neo4j.com/docs/cypher-manual/current/functions/scalar page.

AzuObs:
Thank you both for your valuable input so far!

This pull request is now up for re-review.

Following your initial feedback, and after discussing this feature with Arne and the planner team, there have made major updates to how the PR initial looked:
- I try to keep as much of the wording from the original PR as possible
- I try to keep as many of your suggestions from the original PR as possible
- I try my best to explain path pattern expressions in terms of what they can do, rather than comparing them too much to other features 
- I intentionally don't go into too much detail about "they can be used where a boolean expression is expected" because this would expose a lot of complexity 
- I make some small updates to the examples so they look like they were written by the same person
- I remove the section on the `type()` function which seemed to be out of place

sherfert (approved):
I think this looks good as is!

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -479,33 +479,38 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
-== Using path patterns in `WHERE`
+== Path pattern expressions
 
-[[filter-on-patterns]]
-=== Filter on patterns
+[[path-pattern-expressions]]
+=== About path pattern expressions
+
+Path pattern expressions can be used to assert whether a specified path exists at least once in a graph. Path pattern expressions are similar to xref::subqueries/existential.adoc[existential subqueries].
+On one hand, xref::subqueries/existential.adoc[existential subqueries] are the more powerful feature and can do anything path pattern expressions can, while on the other hand path pattern expressions are more concise.
```
JPryce-Aklundh:
```suggestion
Similar to xref::subqueries/existential.adoc[existential subqueries], path pattern expressions can be used to assert whether a specified path exists at least once in a graph.
While existential subqueries are more powerful and capable of performing anything achievable with path pattern expressions, path pattern expressions are more concise.
```

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -479,33 +479,38 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
 
-Patterns are not only expressions, they are also predicates.
-The only limitation to a pattern is that it must be able to express it in a single path.
-Unlike `MATCH`, it is not possible to use commas between multiple paths.
-To combine multiple patterns using `WHERE`, instead use `AND`.
-Note that new variables cannot be introduced.
-Although it might look very similar to the `MATCH` patterns, the `WHERE` clause is used to eliminate matched paths.
-`MATCH (a)-[*]->(b)` is very different from `WHERE (a)-[*]->(b)`.
-The first will produce a path for every path it can find between `a` and `b`, whereas the latter will eliminate any matched paths where `a` and `b` do not have a directed relationship chain between them.
+Path pattern expressions may only use a subset of xref::patterns/reference.adoc#graph-patterns[graph patterns] semantics.
```
JPryce-Aklundh:
```suggestion
* Path pattern expressions may only use a subset of xref::patterns/reference.adoc#graph-patterns[graph pattern] semantics.
```

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -479,33 +479,38 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
+[[path-pattern-expressions]]
+=== About path pattern expressions
+
+Path pattern expressions can be used to assert whether a specified path exists at least once in a graph. Path pattern expressions are similar to xref::subqueries/existential.adoc[existential subqueries].
+On one hand, xref::subqueries/existential.adoc[existential subqueries] are the more powerful feature and can do anything path pattern expressions can, while on the other hand path pattern expressions are more concise.
 
-Patterns are expressions in Cypher, expressions that return a list of paths.
-List expressions are also predicates; an empty list represents `false`, and a non-empty list represents `true`.
+Path pattern expressions come with several restrictions which are highlighted below.
+Use cases which require more permissive functionality should consider using xref::subqueries/existential.adoc[existential subqueries] instead.
```
JPryce-Aklundh:
```suggestion
Path pattern expressions have the following restrictions (use cases that require extended functionality should consider using xref::subqueries/existential.adoc[existential subqueries] instead):
```

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -479,33 +479,38 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
-The only limitation to a pattern is that it must be able to express it in a single path.
-Unlike `MATCH`, it is not possible to use commas between multiple paths.
-To combine multiple patterns using `WHERE`, instead use `AND`.
-Note that new variables cannot be introduced.
-Although it might look very similar to the `MATCH` patterns, the `WHERE` clause is used to eliminate matched paths.
-`MATCH (a)-[*]->(b)` is very different from `WHERE (a)-[*]->(b)`.
-The first will produce a path for every path it can find between `a` and `b`, whereas the latter will eliminate any matched paths where `a` and `b` do not have a directed relationship chain between them.
+Path pattern expressions may only use a subset of xref::patterns/reference.adoc#graph-patterns[graph patterns] semantics.
+A path pattern expressions must be a xref::patterns/reference.adoc#path-patterns[path pattern] of length greater than zero.
+In other words, it must contain at least one xref::patterns/reference.adoc#relationship-patterns[relationship] or xref::patterns/reference.adoc#variable-length-relationships[variable-length relationship].
```
JPryce-Aklundh:
```suggestion
* A path pattern expression must be a xref::patterns/reference.adoc#path-patterns[path pattern] of length greater than zero.
In other words, it must contain at least one xref::patterns/reference.adoc#relationship-patterns[relationship] or xref::patterns/reference.adoc#variable-length-relationships[variable-length relationship].
```

Should we not link to quantified relationships rather than variable-length relationships (https://neo4j.com/docs/cypher-manual/current/patterns/reference/#quantified-relationships)? With the release of qpps, it was my understanding that we should not promote the old var-length syntax, but I may be missing something here.

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -479,33 +479,38 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
 
+Path pattern expressions may not declare new variables.
+They can only reference existing variables.
+
+Path pattern expressions may only be used in positions where a xref:queries/expressions.adoc#boolean[boolean expression] is expected.
+The following sections will demonstrate how to use path pattern expressions in a `WHERE` clause.
```
JPryce-Aklundh:
```suggestion
* Path pattern expressions may not declare new variables.
They can only reference existing variables.

* Path pattern expressions may only be used in positions where a xref:queries/expressions.adoc#boolean[boolean expression] is expected.
The following sections will demonstrate how to use path pattern expressions in a `WHERE` clause.
```

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -479,33 +479,38 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
 
-
 [[query-where-patterns]]
-== Using path patterns in `WHERE`
+== Path pattern expressions
 
-[[filter-on-patterns]]
-=== Filter on patterns
+[[path-pattern-expressions]]
+=== About path pattern expressions
```
JPryce-Aklundh:
```suggestion
```

JPryce-Aklundh (commented):
Thanks, @AzuObs - this is now much clearer! Let me know what you think of my suggestions.

modules/ROOT/pages/clauses/where.adoc
AzuObs: I could maybe remove the part where I mention the old var-length, if we didn't want to reference that at all? Unfortunately, path pattern expressions cannot use quantified relationships or quantified path patterns.

JPryce-Aklundh: I see. In that case, I think it is better to leave var-lengths in.

modules/ROOT/pages/clauses/where.adoc
```diff
@@ -479,33 +479,36 @@ The `name` and `age` for `Andy` are returned because his name starts with 'AND'
...
 |===
 
-
 [[query-where-patterns]]
-== Using path patterns in `WHERE`
+== Path pattern expressions
 
-[[filter-on-patterns]]
-=== Filter on patterns
+[[path-pattern-expressions]]
```
JPryce-Aklundh, line 482: This anchor should be placed above the header (and replace the old one (query-where-patterns), which ( checked) is currently not used).

```suggestion
```

JPryce-Aklundh (approved):
Apart from one small thing, this looks good! Thanks!

PR #735, opened 2023-09-20
Alter database options

ali-ince:
Adds content for ALTER DATABASE SET/REMOVE OPTION clauses.

TODO: Update links to actual CDC documentation when published.


neo-technology-commit-status-publisher:

ali-ince:
@Hunterness this is a resurrection of https://github.com/neo4j/docs-cypher/pull/286, which is now closed due to my fork being detached.

I addressed your only comment, so hope it's ready for a re-review.

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -903,6 +925,76 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
 [role="statsonlyqueryresult"]
 0 rows
 
+[role=enterprise-edition]
+[[administration-databases-alter-database-options]]
+=== Options
+
+The `ALTER DATABASE` command can be used to set or remove specific options.
```
Hunterness: should definitely mention that this is a subset of the options available for CREATE

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -903,6 +925,76 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
...
+[source, cypher, test-skip]
+----
+SHOW DATABASES yield name, options
+----
+
+.Result
+[role="queryresult",options="header,footer",cols="2*<m"]
+|===
+| +name+ | +options+
+| +"options-example"+ | +{txLogEnrichment: "FULL"}+
```
Hunterness: Either this show example should be before the remove or it feels weird to have anything in the options map. Might actually be nice to do the SET, then SHOW, then REMOVE and then SHOW again to show that it gets removed (or if it falls back to default values then the SET should set something that isn't the default value so we can see the difference)

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -903,6 +925,76 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
...
+|
+Defines level of enrichment applied to transaction logs for Change Data Capture (CDC) purposes.
+
+|===
+
+.+ALTER DATABASE SET OPTION+
+======
+
+.Query
+[source, cypher, test-skip]
```
Hunterness: I guess they have test-skip due to feature flag being off by default? should probably have a comment about why they are skipped anyway (if we can't run them/until we update so they can be run)

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -744,6 +761,11 @@ Defines additional configuration specified by comma separated `name=value` pairs
 |
 Defines credentials that needs to be passed into certain seed providers.
 
+| `txLogEnrichment`
+| `FULL` \| `DIFF` \| `NONE`
+|
+Defines level of enrichment applied to transaction logs for Change Data Capture (CDC) purposes.
```
Hunterness, line 767: do we want to link to CDC docs here?

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -903,6 +925,76 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
...
+
+[options="header"]
+|===
+
+| Key | Value | Description
+
+| `txLogEnrichment`
+| `FULL` \| `DIFF` \| `NONE`
+|
+Defines level of enrichment applied to transaction logs for Change Data Capture (CDC) purposes.
```
Hunterness, line 950: do we want to link to CDC docs here?

modules/ROOT/pages/administration/databases.adoc
ali-ince: Isn't the table below enough?

ali-ince:
back to you @Hunterness :)

modules/ROOT/pages/administration/databases.adoc
Hunterness: While the table shows it is feels nice to point it out explicitly in the text as well (otherwise I'd probably still try setting some of the create thinking they've just been missed)

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -929,13 +930,14 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
...
-There are no available `OPTIONS` values for composite databases.
-====
+////
+[source, cypher, role=test-setup]
+----
+CREATE DATABASE `options-example`;
+----
+////
 
+The `ALTER DATABASE` command can be used to set or remove specific options, which are listed below.
```
Hunterness, line 940:
```suggestion
The `ALTER DATABASE` command can be used to set or remove specific options, which are listed below.
This is a subset of the options allowed for `CREATE DATABASE` as the remaining create options are only applied during the creation of the database.
```
something to that style maybe?

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -1062,7 +1198,8 @@ SHOW DATABASES YIELD name
 | +"sci-fi"+
 | +"system"+
 | +"topology-example"+
-13+d|Rows: 8
+| +"options-example"+
```
Hunterness: I think this list is supposed to be ordered by name so the options one should probably be higher up (even if the docs tests don't check that)

Hunterness (approved):
Just a reminder for the docs team to not miss porting this over to the operations manual since the code is currently duplicated :)

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -903,6 +926,119 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
...
+
+////
+[source, cypher, role=test-setup]
+----
+CREATE DATABASE `options-example`;
+----
+////
+
+The `ALTER DATABASE` command can be used to set or remove specific options, which are listed below.
+This is a subset of the options allowed for `CREATE DATABASE` as the remaining create options are only applied during the creation of the database.
```
stefano-ottolenghi:
```suggestion
```
I don't think this adds any information for the regular user and I'd just take it away.

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -903,6 +926,119 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
...
+
+[NOTE]
+====
+There are no available `OPTIONS` values for composite databases.
+====
+
+.+ALTER DATABASE SET OPTION+
+======
+
+// skip test until we make the change_data_capture feature flag on by default.
```
stefano-ottolenghi, line 963: Do you have an estimate of when this is supposed to happen, if at all? Just to know when we should remember to remove test-skip.

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -903,6 +926,119 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
...
+
+.+SHOW DATABASE+
+======
+
+// skip test until we make the change_data_capture feature flag on by default.
+
+.Query
+[source, cypher, test-skip]
+----
+SHOW DATABASES yield name, options
```
stefano-ottolenghi, line 982:
```suggestion
SHOW DATABASES YIELD name, options
```

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -91,11 +91,24 @@ CREATE OR REPLACE COMPOSITE DATABASE name
...
 {
 SET ACCESS {READ ONLY \| READ WRITE} \|
-SET TOPOLOGY n PRIMAR{Y\|IES} [m SECONDAR{Y\|IES}]
+SET TOPOLOGY n PRIMAR{Y\|IES} [m SECONDAR{Y\|IES}] \|
+SET OPTION option value
 }
 [WAIT [n [SEC[OND[S]]]]\|NOWAIT]
 ----
 
+[source, syntax, role="noheader"]
```
stefano-ottolenghi: FYI `role="noheader"` is not needed anymore. The default for a code block is to _not_ have a header, unless a title is provided with a `.` row preceding the code block.

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -903,6 +926,119 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
...
+There are no available `OPTIONS` values for composite databases.
+====
+
+.+ALTER DATABASE SET OPTION+
+======
+
+// skip test until we make the change_data_capture feature flag on by default.
+
+.Query
+[source, cypher, test-skip]
```
stefano-ottolenghi:
```suggestion
[source, cypher, role="test-skip"]
```
Right now it does work even if you don't put it within the `role` attribute, but I don't want to guarantee that it always will, so please apply this change to further examples below as well.

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -903,6 +926,119 @@ ALTER DATABASE nonExisting IF EXISTS SET TOPOLOGY 1 PRIMARY 0 SECONDARY
...
+
+.+SHOW DATABASE+
+======
+
+// skip test until we make the change_data_capture feature flag on by default.
+
+.Query
+[source, cypher, test-skip]
+----
+SHOW DATABASES yield name, options
```
stefano-ottolenghi:
```suggestion
SHOW DATABASES YIELD name, options
```

renetapopova:
This needs to be updated in the Operations manual -> https://trello.com/c/kAUbbH2p/117-document-alter-database

modules/ROOT/pages/administration/databases.adoc
```diff
@@ -91,11 +91,24 @@ CREATE OR REPLACE COMPOSITE DATABASE name
...
+[source, syntax, role="noheader"]
+----
+ALTER DATABASE name [IF EXISTS]
+REMOVE OPTION option
+[WAIT [n [SEC[OND[S]]]]\|NOWAIT]
+----
+
+[NOTE]
+====
+There can be multiple SET OPTION or REMOVE OPTION clauses for different option keys.
```
stefano-ottolenghi:
```suggestion
There can be multiple `SET OPTION` or `REMOVE OPTION` clauses for different option keys.
```

modules/ROOT/pages/administration/databases.adoc
Hunterness: It comes from one of my review comments. To me if feels strange that there are different things I can add to my create with options and my alter with options, I'd have expected to be able to set the same things in both. This sentence (or some variation of it) does point out to me that it is by design that they are different and that we didn't just miss things for ALTER (or when documenting alter)

modules/ROOT/pages/administration/databases.adoc
Hunterness: I think @l-heemann was looking at enabling it as default in 5.13 but I'm not sure how far he got on that

modules/ROOT/pages/administration/databases.adoc
stefano-ottolenghi: Why would you expect to be able to use on ALTER any other of the options that are available on CREATE? All the other options have to do with seeding the db, or what to do with existing data when the db is first created. None of these would make sense to use on ALTER. 

Whether the options allowed on ALTER are a subset of CREATE or a set with null intersection is un-important to users: either they are the _same_, or they are _different_. Then how much different, and in what way, is not important to the reader. If somebody comes and shouts at us why they can't provide seeding options on ALTER as they can on CREATE, either they don't understood what seeding a db means, or we have failed in explaining it properly.

modules/ROOT/pages/administration/databases.adoc
l-heemann: yes the feature flag is supposed to default true in 5.13 (cdc goes public beta for Nodes).
PR for the feature flag is [here](https://github.com/neo-technology/neo4j/pull/22245) blocked on an issue that is being fixed [here](https://github.com/neo-technology/neo4j/pull/22280)

modules/ROOT/pages/administration/databases.adoc
ali-ince: Removed that line, I think the table just following this paragraph already makes it clear about what options are supported.

stefano-ottolenghi (approved):
This looks good to me! @l-heemann is it possible for you to remember to remove the `test-skip` tags on the relevant examples when CDC is set to true by default, or in alternative to report back here so we remember?

ali-ince:
I'll keep track of it @stefano-ottolenghi.

stefano-ottolenghi:
@ali-ince shall we wait until CDC links are available + flag is turned on, or did you intend to merge this already?

ali-ince:
If we are not publishing these docs until 5.13 release/announcement, I think we can merge - that will also remove the likelihoods of conflicts.

PR #749, opened 2023-09-27
Small fixes

gem-neo4j:
I found some issues with the formatting and links, so fixed them up!

### DIFF
--- a/modules/ROOT/pages/functions/scalar.adoc
+++ b/modules/ROOT/pages/functions/scalar.adoc
@@ -1499,7 +1499,7 @@ With this in mind, the below list contains all supported types (as of Neo4j 5.13
 
 This should be taken into account when relying on the output of the `valueType()` function.
 
-See the xref::syntax/expressions.adoc#type-predicate-expressions[type predicate expression] for an alternative way of testing type values.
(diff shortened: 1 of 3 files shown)


neo-technology-commit-status-publisher:

modules/ROOT/pages/functions/scalar.adoc
```diff
@@ -1499,7 +1499,7 @@ With this in mind, the below list contains all supported types (as of Neo4j 5.13
 
 This should be taken into account when relying on the output of the `valueType()` function.
 
-See the xref::syntax/expressions.adoc#type-predicate-expressions[type predicate expression] for an alternative way of testing type values.
+See the xref::values-and-types/type-predicate.adoc[type predicate expression] for an alternative way of testing type values.
```
JPryce-Aklundh, line 1502: This should be plural as well, right? ("type predicate expressions")

JPryce-Aklundh (approved):
Thanks, @gem-neo4j!

modules/ROOT/pages/functions/scalar.adoc
gem-neo4j: Well it is just one expression though 🤔 Would we say "see the Count Expression" or "see the Count Expressions"? :)

PR #763, opened 2023-10-09
add property rules syntax to keyword glossary

HannesSandberg:


neo-technology-commit-status-publisher:

Hunterness (approved):
We said we left it out of additions/deprecations/removals chapter for now since it's not yet publicly available if I remember correctly.

And I think your check with the docs team yielded only update operations manual privilege chapter.

Based on those assumptions I'll approve this, if any of them is false I guess you know what to do XD

HannesSandberg:
> We said we left it out of additions/deprecations/removals chapter for now since it's not yet publicly available if I remember correctly.

> And I think your check with the docs team yielded only update operations manual privilege chapter.

Yes, that is correct

HannesSandberg:
Closed this PR since this page is being removed and will reference the cheat sheet instead.

Hunterness:
> and will reference the cheat sheet instead.

So then we definitely should have a cheat sheet example for each of the privileges XD (which you did add already)

PR #767, opened 2023-10-13
Add docs for the new nullIf() function

gem-neo4j:
NullIf can be useful for handling special values like 0 or empty strings by replacing them with null, it can also be used in combination with coalesce to substitute the null values with something else. Another use case is to avoid division by zero errors, this is because division by null returns null, whereas division by 0 will error! It is a handy shorthand that SQL users will be familiar with.


neo-technology-commit-status-publisher:

loveleif (approved):
👍 (...except the build is failing for some reason)

PR #775, opened 2023-11-01
Add note for `ORDER BY` and guaranteed row-order

JPryce-Aklundh:
Necessary because of changes that might be coming as a result of parallel runtime.


neo-technology-commit-status-publisher:

PR #781, opened 2023-11-15
Add docs for new NODETACH keyword

gem-neo4j:
Document the new NODETACH keyword

### DIFF
--- a/modules/ROOT/pages/clauses/delete.adoc
+++ b/modules/ROOT/pages/clauses/delete.adoc
@@ -59,7 +59,9 @@ Deleted 1 node
 _This feature was introduced in Neo4j 5.14._
 
 It is also possible to delete the single node using the `NODETACH DELETE` clause.
-This behaves the same as omitting it.
+Using the `NODETACH` keyword explicitly defines that relationships will not be detached and deleted from a node.
+The `NODETACH` keyword is a mirror of the already existing keyword xref:clauses/delete.adoc#delete-a-node-with-all-its-relationships[DETACH],
+Including it is functionally the same as using simple `DELETE`.
 
 .Query
 [source, cypher]
@@ -116,7 +118,6 @@ Deleted 1 node, deleted 1 relationship
 The `DETACH DELETE` clause may not be permitted to users with restricted security privileges. 
 For more information, see link:{neo4j-docs-base-uri}/operations-manual/{page-version}/authentication-authorization/access-control#detach-delete-restricted-user[Operations Manual -> Fine-grained access control].
 ====
-
 
 
 [[delete-all-nodes-and-relationships]]
--- a/modules/ROOT/pages/clauses/delete.adoc
+++ b/modules/ROOT/pages/clauses/delete.adoc
@@ -9,6 +9,7 @@ For removing properties and labels, see the xref::clauses/remove.adoc[REMOVE] cl
 
 It is not possible to delete nodes with relationships connected to them without also deleting the relationships. 
 This can be done by either explicitly deleting specific relationships, or by using the `DETACH DELETE` clause. 
+It is also possible, as of Neo4j 5.14, to explicitly add the keyword `NODETACH`, which has the same behavior as omitting it.
 
 == Example graph 
 
--- a/modules/ROOT/pages/clauses/delete.adoc
+++ b/modules/ROOT/pages/clauses/delete.adoc
@@ -60,7 +60,7 @@ _This feature was introduced in Neo4j 5.14._
 
 It is also possible to delete the single node using the `NODETACH DELETE` clause.
 Using the `NODETACH` keyword explicitly defines that relationships will not be detached and deleted from a node.
-The `NODETACH` keyword is a mirror of the already existing keyword xref:clauses/delete.adoc#delete-a-node-with-all-its-relationships[DETACH],
+The `NODETACH` keyword is a mirror of the already existing keyword xref:clauses/delete.adoc#delete-a-node-with-all-its-relationships[DETACH].
 Including it is functionally the same as using simple `DELETE`.
 
 .Query
--- a/modules/ROOT/pages/clauses/delete.adoc
+++ b/modules/ROOT/pages/clauses/delete.adoc
@@ -9,7 +9,6 @@ For removing properties and labels, see the xref::clauses/remove.adoc[REMOVE] cl
 
 It is not possible to delete nodes with relationships connected to them without also deleting the relationships. 
 This can be done by either explicitly deleting specific relationships, or by using the `DETACH DELETE` clause. 
-It is also possible, as of Neo4j 5.14, to explicitly add the keyword `NODETACH`, which has the same behavior as omitting it.
 
 == Example graph 
 
@@ -53,6 +52,23 @@ This query is only possible to run on nodes without any relationships connected 
 ----
 Deleted 1 node
 ----
+
+[[delete-nodetach]]
+=== NODETACH keyword
+
+_This feature was introduced in Neo4j 5.14._
+
+It is also possible to delete the single node using the `NODETACH DELETE` clause.
+This behaves the same as omitting it.
+
+.Query
+[source, cypher]
+----
+MATCH (n:Person {name: 'Tom Hanks'})
+NODETACH DELETE n
+----
+
+This also deletes the `Person` node `Tom Hanks`.
 
 [[delete-relationships-only]]
 == Delete relationships only
@@ -100,6 +116,7 @@ Deleted 1 node, deleted 1 relationship
 The `DETACH DELETE` clause may not be permitted to users with restricted security privileges. 
 For more information, see link:{neo4j-docs-base-uri}/operations-manual/{page-version}/authentication-authorization/access-control#detach-delete-restricted-user[Operations Manual -> Fine-grained access control].
 ====
+
 
 
 [[delete-all-nodes-and-relationships]]
--- a/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
+++ b/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
@@ -40,6 +40,25 @@ This function returns null if the two given parameters are equivalent, otherwise
 
 |===
 
+=== Updated features
+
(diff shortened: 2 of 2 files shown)


neo-technology-commit-status-publisher:

gem-neo4j:
[GQL compliance - Delete statement](https://trello.com/c/BlynxMe2/765-gql-compliance-delete-statement)

modules/ROOT/pages/clauses/delete.adoc
```diff
@@ -9,6 +9,7 @@ For removing properties and labels, see the xref::clauses/remove.adoc[REMOVE] cl
 
 It is not possible to delete nodes with relationships connected to them without also deleting the relationships. 
 This can be done by either explicitly deleting specific relationships, or by using the `DETACH DELETE` clause. 
+It is also possible, as of Neo4j 5.14, to explicitly add the keyword `NODETACH`, which has the same behavior as omitting it.
```
AzuObs: nitpick: I do not think we should add the Cypher versions in the text, so that the text does not become littered with version numbers for when features began and ended. The docs should simply be an up to date representation of the language. If there is a need for users to refer to specific versions of the docs (say 5.13) then that is a feature that should be handled in a more generic way - until then we should accept that the use case is not met.

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -40,6 +40,25 @@ This function returns null if the two given parameters are equivalent, otherwise
...
+=== Updated features
+
+[cols="2", options="header"]
+|===
+| Feature
+| Details
+
+a|
+label:functionality[]
+label:updated[]
```
AzuObs: nitpick: What is the reason why this is not considered an addition?

AzuObs (approved):
LGTM 👍

modules/ROOT/pages/clauses/delete.adoc
gem-neo4j: Yeah, I will see what @JPryce-Aklundh says as we have been adding a bit more versioning in the docs :) but probably a better way to do it for sure!

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
gem-neo4j: I just felt like it was an extension of the the DELETE clause, more so than a standalone addition, also the functionality added is not new, so I felt like it was strange to imply as such, but not really bothered by which, so can change if you like

modules/ROOT/pages/clauses/delete.adoc
JPryce-Aklundh: When a new feature is released we handle it with both labels and more formal "this feature was released in ..." notes. 
For this case, maybe instead of mentioning it in the introductory paragraph, we can add a subsection under the current "Delete single node" header called "NODETACH", and this can begin with the standard "This feature was introduced in Neo4j 5.14." note - such a section could also include an example of how to use NODETACH - which I think is currently missing.

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -40,6 +40,25 @@ This function returns null if the two given parameters are equivalent, otherwise
...
+[source, cypher, role="noheader"]
+----
+MATCH (n) NODETACH DELETE n
+----
+a|
+Added a new keyword `NODETACH`, explicitly defining that relationships will not be detached and deleted.
+This is a mirror of the already existing keyword `DETACH`.
```
JPryce-Aklundh, line 59: We have used "alias" elsewhere to describe these features - would that work here instead of mirror?

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
JPryce-Aklundh: I'm also inclined to think this should be labelled as a new feature

modules/ROOT/pages/clauses/delete.adoc
```diff
@@ -9,6 +9,7 @@ For removing properties and labels, see the xref::clauses/remove.adoc[REMOVE] cl
 
 It is not possible to delete nodes with relationships connected to them without also deleting the relationships. 
 This can be done by either explicitly deleting specific relationships, or by using the `DETACH DELETE` clause. 
+It is also possible to explicitly add the keyword `NODETACH`, which has the same behavior as omitting it.
```
JPryce-Aklundh: I would probably remove mentioning it here to avoid unnecessary confusion

```suggestion
```

modules/ROOT/pages/clauses/delete.adoc
```diff
@@ -53,6 +54,28 @@ This query is only possible to run on nodes without any relationships connected
...
 
+=== NODETACH keyword
+
+_This feature was introduced in Neo4j 5.14._
+
+It is also possible to delete the single node using the `NODETACH DELETE` clause.
+This behaves the same as omitting it.
+
+.Query
+[source, cypher, indent=0]
```
JPryce-Aklundh: We don't need the indent=0 anymore (I have been meaning to get rid of all existing examples but havent yet)

```suggestion
[source, cypher]
```

modules/ROOT/pages/clauses/delete.adoc
```diff
@@ -101,6 +124,33 @@ For more information, see link:{neo4j-docs-base-uri}/operations-manual/{page-ver
...
+Deleted 1 node, deleted 1 relationship
+----
+
+[NOTE]
+====
+The `DETACH DELETE` clause may not be permitted to users with restricted security privileges.
+For more information, see link:{neo4j-docs-base-uri}/operations-manual/{page-version}/authentication-authorization/access-control#detach-delete-restricted-user[Operations Manual -> Fine-grained access control].
+====
+
+
```
JPryce-Aklundh: This part is slightly confusing for two reasons:
- it creates a new header called "Delete single node" (there is already one with the same name - see line 36)
- it first states "delete a single node that has no relationships connected to it.." and then, after the example, says "This deletes the `Person` node `Carrie-Anne Moss` and all relationships connected to it." (It may not be obvious to readers that Carrie's relationship was deleted in the previous example).

More broadly, I don't think this part needs to be included. We have already mentioned NODETACH above, which probably includes all that is necessary, right?

modules/ROOT/pages/clauses/delete.adoc
```diff
@@ -53,6 +54,28 @@ This query is only possible to run on nodes without any relationships connected
 Deleted 1 node
 ----
 
+=== NODETACH keyword
+
+_This feature was introduced in Neo4j 5.14._
+
+It is also possible to delete the single node using the `NODETACH DELETE` clause.
+This behaves the same as omitting it.
```
JPryce-Aklundh: This is very tricky, and I think we need to be more explicit (at least until we can mention GQL, which will make documenting these features so much simpler)

```suggestion
Using the `NODETACH` keyword explicitly defines that relationships will not be detached and deleted from a node.
The `NODETACH` keyword is a mirror of the already existing keyword xref:clauses/delete.adoc#delete-a-node-with-all-its-relationships[DETACH], 
Including it is functionally the same as using simple `DELETE`.
```

modules/ROOT/pages/clauses/delete.adoc
```diff
@@ -53,6 +54,28 @@ This query is only possible to run on nodes without any relationships connected
...
+----
+
+This also deletes the `Person` node `Tom Hanks`.
+
+.Result
+[role="queryresult",options="footer",cols="1*<m"]
+----
+Deleted 1 node
+----
+
```
JPryce-Aklundh:
```suggestion
```

Maybe confusing to readers if we add results of deleting the same node twice.

modules/ROOT/pages/clauses/delete.adoc
```diff
@@ -53,6 +54,28 @@ This query is only possible to run on nodes without any relationships connected
 Deleted 1 node
 ----
 
+=== NODETACH keyword
```
JPryce-Aklundh, line 57:
```suggestion
[delete-nodetach]
=== NODETACH keyword
```

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -38,6 +38,16 @@ RETURN nullIf(v1, v2)
...
 
+a|
+label:functionality[]
+label:new[]
+[source, cypher, role="noheader"]
+----
+MATCH (n) NODETACH DELETE n
+----
+a|
+Added a new keyword `NODETACH`, explicitly defining that relationships will not be detached and deleted.
```
JPryce-Aklundh: adds link to location in docs using new anchor created in suggestion above.

```suggestion
Added a new keyword xref:clauses/delete.adoc#delete-nodetach[NODETACH], explicitly defining that relationships will not be detached and deleted.
```

modules/ROOT/pages/clauses/delete.adoc
gem-neo4j: Oh this wasn't meant to be here 🙈 I forgot to remove it

modules/ROOT/pages/clauses/delete.adoc
```diff
@@ -53,6 +53,25 @@ This query is only possible to run on nodes without any relationships connected
...
 ----
 
+[[delete-nodetach]]
+=== NODETACH keyword
+
+_This feature was introduced in Neo4j 5.14._
+
+It is also possible to delete the single node using the `NODETACH DELETE` clause.
+Using the `NODETACH` keyword explicitly defines that relationships will not be detached and deleted from a node.
+The `NODETACH` keyword is a mirror of the already existing keyword xref:clauses/delete.adoc#delete-a-node-with-all-its-relationships[DETACH],
```
JPryce-Aklundh:
```suggestion
The `NODETACH` keyword is a mirror of the already existing keyword xref:clauses/delete.adoc#delete-a-node-with-all-its-relationships[DETACH].
```

JPryce-Aklundh (approved):
Looks good! Thanks!

PR #794, opened 2023-11-23
Update the links to syntax description section

Hunterness:
Since they were broken after it moved to the operations manual.

Also add a comment around the new vector index type where we explain different index types same as how it looks for full-text indexes.

### DIFF
--- a/modules/ROOT/pages/clauses/listing-functions.adoc
+++ b/modules/ROOT/pages/clauses/listing-functions.adoc
@@ -76,7 +76,7 @@ m| BOOLEAN
 
 [NOTE]
 ====
-More details about the syntax descriptions can be found xref:administration/index.adoc#administration-syntax[here].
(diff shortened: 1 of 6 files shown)


neo-technology-commit-status-publisher:

modules/ROOT/pages/clauses/listing-functions.adoc
```diff
@@ -76,7 +76,7 @@ m| BOOLEAN
 
 [NOTE]
 ====
-More details about the syntax descriptions can be found xref:administration/index.adoc#administration-syntax[here].
+More details about the syntax descriptions can be found link:{neo4j-docs-base-uri}/operations-manual/{page-version}/database-administration/syntax.adoc#administration-syntax-reading[here].
```
Hunterness: and apparently that isn't the correct link either (gives page not found in preview)

Hunterness: And fixed

AlexicaWright (approved):

PR #795, opened 2023-11-24
Update the links to syntax description section (#794)

AlexicaWright:
Since they were broken after it moved to the operations manual.

Also add a comment around the new vector index type where we explain different index types same as how it looks for full-text indexes.


neo-technology-commit-status-publisher:
This PR includes documentation updates.

You can view the updated docs at https://neo4j-docs-cypher-795.surge.sh

PR #810, opened 2023-12-13
Allow parameters for index and constraint names in create and drop commands

Hunterness:


neo-technology-commit-status-publisher:

modules/ROOT/pages/constraints/examples.adoc
```diff
@@ -52,6 +53,45 @@ It will be updated to say `Node property uniqueness constraints added: 1` in Neo
...
+The constraint name can also be given as a parameter.
+
+.+CREATE CONSTRAINT+
+======
+
+.Parameters
+[source,javascript, indent=0]
+----
+{
+  "name": "node_uniqueness_param"
```
l-heemann, line 69: Since we are placing this example as a contrast to the previous example. I think it would make sense to use the same property and parameter name.

l-heemann: Same for the other examples

modules/ROOT/pages/constraints/examples.adoc
```diff
@@ -2555,50 +2885,64 @@ SHOW CONSTRAINTS
...
+├───┼──────────────────────────────┼─────────────────────────────────┼──────────────┼─────────────┼──────────────────────────────────┼──────────────────────────────┼────────────────────────────────┤
+│27 │"wrote_locations"             │"RELATIONSHIP_PROPERTY_EXISTENCE"│"RELATIONSHIP"│["WROTE"]    │["location"]                      │null                          │null                            │
+├───┼──────────────────────────────┼─────────────────────────────────┼──────────────┼─────────────┼──────────────────────────────────┼──────────────────────────────┼────────────────────────────────┤
+│25 │"wrote_year"                  │"RELATIONSHIP_PROPERTY_EXISTENCE"│"RELATIONSHIP"│["WROTE"]    │["year"]                          │null                          │null                            │
+└───┴──────────────────────────────┴─────────────────────────────────┴──────────────┴─────────────┴──────────────────────────────────┴──────────────────────────────┴────────────────────────────────┘
+27 rows
```
l-heemann, line 2919: This looks terrible non-monospaced. Asked docs for help https://neo4j.slack.com/archives/C02JR5E4L/p1702641273166879

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -463,11 +526,20 @@ Only one relationship type lookup index can exist at a time.
 ===== Create a token lookup index only if it does not already exist
 
 If it is not known whether an index exists or not, add `IF NOT EXISTS` to ensure it does.
+Giving the name directly or through a parameter makes no difference.
```
l-heemann: I don't think we need to write this. This is sensible enough that it should be expected. Mentioning this just causes the reader to question whether there are cases where parameter vs literal would make a difference.

l-heemann (changes requested):

modules/ROOT/pages/constraints/examples.adoc
Hunterness: that would fail docs tests though, I can't create the same index twice...

modules/ROOT/pages/indexes-for-search-performance.adoc
Hunterness: I think I added it here just cause I couldn't add a new lookup index example, so therefore I also added the comment but I can remove the comment

modules/ROOT/pages/constraints/examples.adoc
Hunterness: or I'm I misunderstanding what you asked for to begin with?

Hunterness: After discussions we decided it was better to keep as is and show them down in the show examples than duplicating the above examples index and skip in testing

modules/ROOT/pages/constraints/examples.adoc
Hunterness: should hopefully be fixed now

modules/ROOT/pages/constraints/examples.adoc
```diff
@@ -52,6 +53,45 @@ It will be updated to say `Node property uniqueness constraints added: 1` in Neo
...
+.Result
+[queryresult]
+----
+Added 1 constraint.
+----
+
+[NOTE]
+====
+The detailed statistics view currently says `Unique constraints added: 1`.
+It will be updated to say `Node property uniqueness constraints added: 1` in Neo4j version 6.0.
```
JPryce-Aklundh: I wonder if this is something users will care about, and therefore something we have to communicate? Or is it necessary because it constitutes a breaking change?

```suggestion
It will be updated to say `Node property uniqueness constraints added: 1` in Neo4j 6.0.
```
Same for all instances of "version".

l-heemann (approved):

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -247,6 +269,26 @@ The following statement will create a named text index on all relationships with
 CREATE TEXT INDEX rel_text_index_name FOR ()-[r:KNOWS]-() ON (r.interest)
 ----
 
+[discrete]
+[[indexes-create-a-text-index-by-param]]
+===== Create a text index using a parameter
+
+The following statement will create a named text index on all nodes labeled with `Person` and which have the `favoriteColor` string property using a parameter for the index name.
```
JPryce-Aklundh:
```suggestion
The following statement will create a named text index on all nodes labeled with `Person` and which have the `favoriteColor` `STRING` property using a parameter for the index name.
```

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -341,6 +384,26 @@ The following statement will create a named point index on all relationships wit
 CREATE POINT INDEX rel_point_index_name FOR ()-[r:STREET]-() ON (r.intersection)
 ----
 
+[discrete]
+[[indexes-create-a-point-index-by-param]]
+===== Create a point index using a parameter
+
+The following statement will create a named point index on all relationships with relationship type `STREET` and point property `coordinate` using a parameter for the index name.
```
JPryce-Aklundh:
```suggestion
The following statement will create a named point index on all relationships with relationship type `STREET` and `POINT` property `coordinate` using a parameter for the index name.
```

JPryce-Aklundh (commented):
Looks good overall. I just have a question about the need to notify readers about the status message change in 6.0 (and if yes, whether we need to mention it under example), and a few suggestions to update formating of data types.

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -170,6 +171,26 @@ The following statement will create a named composite range index on all relatio
 CREATE INDEX composite_range_rel_index_name FOR ()-[r:PURCHASED]-() ON (r.date, r.amount)
 ----
 
+[discrete]
+[[indexes-create-a-range-index-by-param]]
+===== Create a range index using a parameter
+
```
JPryce-Aklundh, line 177:
```suggestion
===== Create a range index using a parameter

_This feature was introduced in Neo4j 5.16._
```

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -247,6 +269,26 @@ The following statement will create a named text index on all relationships with
 CREATE TEXT INDEX rel_text_index_name FOR ()-[r:KNOWS]-() ON (r.interest)
 ----
 
+[discrete]
+[[indexes-create-a-text-index-by-param]]
+===== Create a text index using a parameter
+
```
JPryce-Aklundh, line 277:
```suggestion
===== Create a text index using a parameter

_This feature was introduced in Neo4j 5.16._

```

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -341,6 +384,26 @@ The following statement will create a named point index on all relationships wit
 CREATE POINT INDEX rel_point_index_name FOR ()-[r:STREET]-() ON (r.intersection)
 ----
 
+[discrete]
+[[indexes-create-a-point-index-by-param]]
+===== Create a point index using a parameter
+
```
JPryce-Aklundh, line 394:
```suggestion
===== Create a point index using a parameter

_This feature was introduced in Neo4j 5.16._

```

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -833,6 +909,29 @@ DROP INDEX example_index
 If an index with that name exists it is removed, if not the command fails.
 
 
+[discrete]
+[[indexes-drop-an-index-by-param]]
+==== Drop an index using a parameter
+
```
JPryce-Aklundh, line 921:
```suggestion
==== Drop an index using a parameter

_This feature was introduced in Neo4j 5.16._

```

modules/ROOT/pages/indexes-for-full-text-search.adoc
```diff
@@ -222,10 +224,18 @@ This means that updates will be applied in a background thread "as soon as possi
...
 ======
 
+.Parameters
+[source,javascript, indent=0]
+----
+{
+  "name": "taggedByRelationshipIndex"
+}
+----
+
```
JPryce-Aklundh, line 236: Should the only create full-text index be with using a param?
Couldn't we keep this example as is, and then say something like. 
"As of Neo4j 5.16, it is also possible to create a named full-text index using parameters."  and then have an example of this?

modules/ROOT/pages/indexes-for-vector-search.adoc
```diff
@@ -113,8 +113,10 @@ The options map is mandatory because setting the vector dimensions and similarit
 
 |===
 
-It is considered best practice to give the index a name when it is created. This name is needed for both dropping and querying the index.
+It is considered best practice to give the index a name when it is created.
+This name is needed for both dropping and querying the index.
 If the index is not explicitly named, it will get an auto-generated name.
+The index name can also be given as a parameter, `CREATE VECTOR INDEX $name FOR ...`.
```
```suggestion
As of Neo4j 5.16, the index name can also be given as a parameter, `CREATE VECTOR INDEX $name FOR ...`.
```

modules/ROOT/pages/constraints/examples.adoc
Hunterness: The comment was added cause it was decided it was to breaking to just update (not sure if all it included the statistics messages as well or just all other places but it was decided to not update any of them now and do all of them later since at least most places (like index type and error messages) were definitely considered breaking.

I think that when the comments were first added the examples actually printed the original string, but they have since then been updated not to, which makes the comments feel like they miss context 🤷

modules/ROOT/pages/indexes-for-search-performance.adoc
Hunterness: this and the next comment would need to be updated for all text and point examples not just the new ones I believe

modules/ROOT/pages/indexes-for-full-text-search.adoc
Hunterness: We have the node example which doesn't use parameter, so we do have one of each. It didn't feel like it matched the page outline to add another example like how it's done for the range/text/point

modules/ROOT/pages/indexes-for-search-performance.adoc
JPryce-Aklundh: The formatting will all be fixed in my new chapter

modules/ROOT/pages/indexes-for-full-text-search.adoc
JPryce-Aklundh: Ok. Maybe at the start of the example box, we can say something like
"This statement creates a named full-text index using a parameter. Creating and dropping indexes using parameters was introduced in Neo4j 5.16."

And then I can see if I can make it fit in a more neater way in my new full-text index page.

JPryce-Aklundh (approved):

modules/ROOT/pages/indexes-for-search-performance.adoc
```diff
@@ -170,6 +171,27 @@ The following statement will create a named composite range index on all relatio
 CREATE INDEX composite_range_rel_index_name FOR ()-[r:PURCHASED]-() ON (r.date, r.amount)
 ----
 
+[discrete]
+[[indexes-create-a-range-index-by-param]]
+===== Create a range index using a parameter
+
+_This feature was introduced in Neo4j 5.16._
+The following statement will create a named range index on all nodes labeled with `Person` and which have the `firstname` property using a parameter for the index name.
```
JPryce-Aklundh:
```suggestion
_This feature was introduced in Neo4j 5.16._

The following statement will create a named range index on all nodes labeled with `Person` and which have the `firstname` property using a parameter for the index name.
```

PR #806, opened 2023-12-07
Add more details as to why a user would want to use nullIf

gem-neo4j:
There has been some confusion as to the use case of nullIf, this adds a bit more reasoning in the docs as to how a user might use it!

PR #818, opened 2023-12-20
Add more details as to why a user would want to use nullIf (#806)

JPryce-Aklundh:
There has been some confusion as to the use case of nullIf, this adds a bit more reasoning in the docs as to how a user might use it!

---------


neo-technology-commit-status-publisher:

PR #838, opened 2024-01-15
Convert Constraint Types to headers (#837)

JPryce-Aklundh:
This would allow us to link users to a specific type of constraint.

---------


neo-technology-commit-status-publisher:

PR #885, opened 2024-02-07
Add the genai.vector.encode function to the available functions

renetapopova:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-885.surge.sh

Updated pages:
- [Functions](https://neo4j-docs-cypher-885.surge.sh/cypher-manual/5/functions)

neo-technology-commit-status-publisher:

modules/ROOT/pages/functions/index.adoc
```diff
@@ -797,6 +798,18 @@ Database functions provide information about databases.
 label:new[Introduced in 5.12]
 |===
 
+[[header-query-functions-genai]]
+**xref::genai-integrations.adoc#single-embedding[GenAI function]** label:new[Introduced in 5.16] label:aura-only[Available only on Aura]
```
nilsceberg, line 802: The heading above says "functions" in the plural even though that also only has one function.

modules/ROOT/pages/functions/index.adoc
```diff
@@ -797,6 +798,18 @@ Database functions provide information about databases.
...
+
+This function is used to generate a vector embedding for a single value.
```
nilsceberg, line 804: Should there be a link to the plugin docs?

nilsceberg (approved):
This looks good, but what do we do when the plugin is released for on-prem users? Simply change "Available only on Aura" to "Requires GenAI plugin" or something?

modules/ROOT/pages/functions/index.adoc
renetapopova: This link is to the plugin docs.

modules/ROOT/pages/functions/index.adoc
renetapopova: Do you mean the anchor?

renetapopova:
> This looks good, but what do we do when the plugin is released for on-prem users? Simply change "Available only on Aura" to "Requires GenAI plugin" or something?

We could remove the label "Available only on Aura" and add a note that for on-prem they need to install the plugin.

modules/ROOT/pages/functions/index.adoc
renetapopova: I mean `xref::genai-integrations.adoc#single-embedding`

nilsceberg: Oh, sorry, I'm just blind!

modules/ROOT/pages/functions/index.adoc
nilsceberg: No, I meant that this heading says "GenAI function" but the one above says "Database functions", even though there is only one function under that heading as well. Seemed like they should be consistent in that regard. :)

PR #901, opened 2024-03-04
Fix CASE example

gem-neo4j:
There was an ambiguous case we had to remove in CASE expressions.

### DIFF
--- a/modules/ROOT/pages/queries/case.adoc
+++ b/modules/ROOT/pages/queries/case.adoc
@@ -107,9 +107,9 @@ The supported comparators are:
 
 * xref::syntax/operators.adoc#query-operators-comparison[Regular Comparison Operators]: `+=+`, `+<>+`, `+<+`, `+>+`, `+<=+`, `+>=+`
 * xref:values-and-types/working-with-null.adoc#is-null-is-not-null[`IS NULL` Operator]: `IS [NOT] NULL`
-* xref:values-and-types/type-predicate.adoc[Type Predicate Expression]: `IS [NOT] :: <TYPE>`
+* xref:values-and-types/type-predicate.adoc[Type Predicate Expression]: `IS [NOT] TYPED <TYPE>` (Note: the form `IS [NOT] :: <TYPE>` is not currently accepted as it leads to an ambiguous grammar)
 * xref::syntax/operators.adoc#match-string-is-normalized[Normalization Predicate Expression]: `IS [NOT] NORMALIZED`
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-901.surge.sh

Updated pages:
- [Conditional expressions (CASE)](https://neo4j-docs-cypher-901.surge.sh/cypher-manual/5/queries/case)

modules/ROOT/pages/queries/case.adoc
```diff
@@ -107,9 +107,9 @@ The supported comparators are:
 
 * xref::syntax/operators.adoc#query-operators-comparison[Regular Comparison Operators]: `+=+`, `+<>+`, `+<+`, `+>+`, `+<=+`, `+>=+`
 * xref:values-and-types/working-with-null.adoc#is-null-is-not-null[`IS NULL` Operator]: `IS [NOT] NULL`
-* xref:values-and-types/type-predicate.adoc[Type Predicate Expression]: `IS [NOT] :: <TYPE>`
+* xref:values-and-types/type-predicate.adoc[Type Predicate Expression]: `IS [NOT] TYPED <TYPE>` (Note: the form `IS [NOT] :: <TYPE>` is not currently accepted as it leads to an ambiguous grammar)
```
rsill-neo4j:
```suggestion
* xref:values-and-types/type-predicate.adoc[Type Predicate Expression]: `IS [NOT] TYPED <TYPE>` (Note that the form `IS [NOT] :: <TYPE>` is currently not accepted as it leads to ambiguous grammar)
```

what is the ambiguity? would it make sense to add an example?

gem-neo4j: I think an example might be more confusing to users, it was a very niche edge case:

WITH "this is a string" AS IS
RETURN CASE 1
    WHEN IS :: INT THEN "1 is an int"
    ELSE "the IS is actually a variable so this is returned"
END

JPryce-Aklundh: I don't think we need to give a reason here, and we also try to avoid words like "currently". In other words, can't we just say "Note that the form... is not accepted", and then change it in the future, if and when it becomes necessary?

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-901.surge.sh

Updated pages:
- [Conditional expressions (CASE)](https://neo4j-docs-cypher-901.surge.sh/cypher-manual/5/queries/case)

JPryce-Aklundh (approved):

rsill-neo4j (approved):

neo-technology-commit-status-publisher:

PR #899, opened 2024-02-28
Update comment about all available types under list of allowed types for property type constraints

Hunterness:
Was a bit confusing to first have supported types and then say available types (as available could be seen as 'available to the constraints as well')

PR #902, opened 2024-03-04
Update comment about all available types under list of allowed types for property type constraints (#899)

JPryce-Aklundh:
Was a bit confusing to first have supported types and then say available types (as available could be seen as 'available to the constraints as well')


neo-technology-commit-status-publisher:

PR #912, opened 2024-03-05
Document AzureOpenAI and dimensions parameter support in GenAI plugin

nilsceberg:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-912.surge.sh

Updated pages:
- [GenAI integrations](https://neo4j-docs-cypher-912.surge.sh/cypher-manual/5/genai-integrations)

parnmatt:
L19, I understand it is a list of examples, but it was extensive before. Should it also now note an Azure account?

modules/ROOT/pages/genai-integrations.adoc
```diff
@@ -266,9 +266,46 @@ Supported values:
...
+
+| dimensions
+| INTEGER
+| The number of dimensions you want to reduce the vector to. Only supported for certain models (see official documentation).
+| Model-dependent.
+|===
+
+=== Azure OpenAI
+
+* Identifier: `"AzureOpenAI"`
```
parnmatt: we call it `Identifier` here (and for the rest) … thinking about this, shouldn't this note that this is the "provider name" to tied to more closely to the interface?

"provider name identifier" perhaps?

nilsceberg: Yeah honestly I wasn't sure if this was clear enough when I wrote it originally. I tried to clarify by documenting the `provider` parameter as "the case-insensitive identifier of the provider to use". Not a huge fan of "provider name identifier" ("name identifier" seems a bit redundant) but maybe just making it very explicit that it is the corresponding argument for the `provider` parameter?

parnmatt: yeah it's a bit much, what do you have in mind for the explicit note?

nilsceberg: Something like
```suggestion
* Identifier (`provider` argument): `"AzureOpenAI"`
```
perhaps?

parnmatt: Hmmnnn, seems a little odd, what not just
```adoc
`provider` argument: `'AzureOpenAI'`
```

nilsceberg: I'd argue that still calling it "identifier" is useful to match the documentation of the parameter. I imagined "`provider` argument" as just clarification of where it's applicable :)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-912.surge.sh

Updated pages:
- [GenAI integrations](https://neo4j-docs-cypher-912.surge.sh/cypher-manual/5/genai-integrations)

parnmatt (approved):

modules/ROOT/pages/genai-integrations.adoc
```diff
@@ -266,19 +266,56 @@ Supported values:
...
 | STRING
-| The ID of the model you want to invoke. +
-Supported values: `"text-embedding-ada-002"`
-| "text-embedding-ada-002"
+| The name of the model you want to invoke.
+| `"text-embedding-ada-002"`
+
+| dimensions
+| INTEGER
+| The number of dimensions you want to reduce the vector to. Only supported for certain models (see official documentation).
```
JPryce-Aklundh: I would either be more specific with regard to which models this applies to, or delete the bracketed content - which isnt telling the user a lot. 

```suggestion
| The number of dimensions you want to reduce the vector to. Only supported for certain models.
```

modules/ROOT/pages/genai-integrations.adoc
```diff
@@ -266,19 +266,56 @@ Supported values:
...
+| Model-dependent.
+|===
+
+=== Azure OpenAI
+
+* Identifier (`provider` argument): `"AzureOpenAI"`
+* https://learn.microsoft.com/en-us/azure/ai-services/openai/[Official Azure OpenAI documentation]
+
+Not that unlike the other providers, the model is configured when creating the deployment on Azure, and is thus not part of the configuration map.
```
JPryce-Aklundh:
```suggestion
Note that, unlike the other providers, the model is configured when creating the deployment on Azure, and is thus not part of the configuration map.
```

modules/ROOT/pages/genai-integrations.adoc
```diff
@@ -266,19 +266,56 @@ Supported values:
...
+| label:required[]
+
+| deployment
+| STRING
+| The name of the model deployment.
+| label:required[]
+
+| dimensions
+| INTEGER
+| The number of dimensions you want to reduce the vector to. Only supported for certain models (see official documentation).
```
JPryce-Aklundh: See above comment.

```suggestion
| The number of dimensions you want to reduce the vector to. Only supported for certain models.
```

modules/ROOT/pages/genai-integrations.adoc
```diff
@@ -266,19 +266,56 @@ Supported values:
...
+| The name of the model you want to invoke.
+| `"text-embedding-ada-002"`
+
+| dimensions
+| INTEGER
+| The number of dimensions you want to reduce the vector to. Only supported for certain models (see official documentation).
+| Model-dependent.
+|===
+
+=== Azure OpenAI
```
JPryce-Aklundh, line 279: This will add a label aligned with the header specifying the release version.
```suggestion
[role=label--new-5.18]
=== Azure OpenAI
```

JPryce-Aklundh (commented):
Thanks @nilsceberg  - looks good overall, just a few suggestions from me. Also, I think we should probably add something to the Additions page, saying that AzureOpenAI is supported.

modules/ROOT/pages/genai-integrations.adoc
parnmatt: Well we may have to be careful here, should it be `5.18` (Neo4j version its packaged with) or `1.1.0` (the plugin version).
This is something we've been debating in other areas.

JPryce-Aklundh: Ok - I can't speak to the plugin version, but if a user can only use it on 5.18 and later, then this should be added. Could we mention the plugin version details in the main body of the text?

parnmatt: the plugin _was_ independent of the neo4j version; we've since added some private metrics which _does_ effectively tie it to a minimum neo4j version; but I believe we've just noted that it is from Neo4j 5.17 as a 'release' (despite actually being packaged earlier)

modules/ROOT/pages/genai-integrations.adoc
nilsceberg: I'd like to avoid being too specific as we don't have any control of any changes being made on the OpenAI side, and I think this page should avoid (in general) documenting the providers rather than the plugin (but that line is blurry at times).

Would a direct link to the relevant part of the OpenAI docs be better?

JPryce-Aklundh: Either that, or just simply "Only supported for certain models."

nilsceberg: All righty 👍

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-912.surge.sh

Updated pages:
- [GenAI integrations](https://neo4j-docs-cypher-912.surge.sh/cypher-manual/5/genai-integrations)

modules/ROOT/pages/genai-integrations.adoc
nilsceberg: Since I think the conclusion from the other discussions was that it's reasonable to make the plugin version the same as the Neo4j version (but hide it in Aura), I think this label makes sense.

JPryce-Aklundh (approved):

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-912.surge.sh

Updated pages:
- [GenAI integrations](https://neo4j-docs-cypher-912.surge.sh/cypher-manual/5/genai-integrations)

neo-technology-commit-status-publisher:

PR #954, opened 2024-05-07
New vector index page

JPryce-Aklundh:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
-| `+CREATE VECTOR INDEX ...+`
-| Create a vector index for the specified label and property with the given vector dimensionality using the given similarity function.
-See the xref:indexes/search-performance-indexes/managing-indexes.adoc#create-indexes[`CREATE INDEX`] command for more details. Replaces {link-procedures-reference}#procedure_db_index_vector_createNodeIndex[`db.index.vector.createNodeIndex`]. label:new[Introduced in 5.15] 
+The graph contains 28863 nodes and 332522 relationships.
 
-| Create relationship vector index.
-| `+CREATE VECTOR INDEX ...+`
-| Create a relationship vector index for the specified relationship type and property with the given vector dimensionality using the given similarity function.
-See the xref:indexes/search-performance-indexes/managing-indexes.adoc#create-indexes[`CREATE INDEX`] command for more details. label:new[Introduced in 5.18]
+To recreate the graph, download and import the link:https://github.com/neo4j-graph-examples/recommendations/blob/main/data/recommendations-embeddings-50.dump[5.0 dump file] to an empty Neo4j database.
```
parnmatt: This dump file appears to be a 5.18 dump file, thus would require Neo4j 5.18+ to read it.

Whoever created it probably used a 5.0 database dump as the source, but then used Neo4j 5.18 to write the embeddings into the database. If we want this dump file to still work on 5.0 it will have to be recreated using only neo4j 5.0.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
-== Vector index commands and procedures
+Neo4j vector indexes are powered by the link:https://lucene.apache.org/[Apache Lucene] indexing and search library.footnote:[Lucene implements a Hierarchical Navigable Small World (HNSW) Graph to perform a k approximate nearest neighbors (k-ANN) query over the vector fields.
+For more information, see link:http://dx.doi.org/10.1109/TPAMI.2018.2889473[Efficient and Robust Approximate Nearest Neighbor Search Using Hierarchical Navigable Small World Graphs] -- Yury A. Malkov and Dmitry A. Yashunin]
 
-Vector indexes are managed through Cypher commands and built-in procedures, see {link-procedures-reference}[Operations Manual -> Procedures] for a complete reference.
+== Example graph
 
-The procedures and commands for vector indexes are listed in the following table:
+The examples on this page use the link:https://github.com/neo4j-graph-examples/recommendations[Neo4j movie recommendations] dataset, focusing on the `plot` and `embedding` properties of `Movie` nodes.
+The `embedding` property consists of a 1536-dimension vector embedding of the `plot` and `title` property combined.
```
parnmatt, line 25: may want to note where these embeddings come from such that a user can use the same model if they are extending this dataset, and/or querying. Considering it's 1536d, it's almost certainly OpenAI's `text-embedding-ada-002`, though should be confirmed with the creator of the data.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
-| {link-procedures-reference}#procedure_db_index_vector_queryNodes[`db.index.vector.queryNodes`]
-| Query the given node vector index.
-Returns the requested number of approximate nearest neighbor nodes and their similarity score, ordered by score.
-
-| Use relationship vector index.
-| {link-procedures-reference}#procedure_db_index_vector_queryRelationships[`db.index.vector.queryRelationships`]
-| Query the given relationship vector index.
-Returns the requested number of approximate nearest neighbor relationships and their similarity score, ordered by score. label:new[Introduced in 5.18]
+[NOTE]
+The dump file used to load the dataset contains embeddings generated by https://openai.com/[OpenAI].
```
parnmatt: Oh, I see it is here. Ok, then the note of the _model_ should be here then.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
...6108556669205427, -0.0008832658641040325, -0.0267605260014534, 0.02873126231133938, -0.022935740649700165, 0.02412855438888073, -0.005513523705303669, 0.00507918419316411, 0.005951105151325464, 0.005173183046281338, 0.005001391749829054, -0.0177625585347414, 0.015130587853491306, -0.01999260112643242, -0.0192406103014946, 0.02140582725405693, -0.007377295754849911, 0.02927580662071705, -0.015104657039046288, -0.01216800045222044, 0.0034552712459117174, -0.003481202060356736, 0.013561777770519257, 0.0022592158056795597, -0.01183090079575777, 0.013665501028299332, 0.017075393348932266, 0.008511765860021114, -0.007409709505736828, -0.0014261907199397683, -0.003058206755667925, 0.0031424816697835922, 0.0021846650633960962, -0.000989419873803854, -0.022144854068756104, 0.0036270625423640013, -0.006233102176338434, -0.016504917293787003, 0.003999816719442606, 0.017931107431650162, -0.025956671684980392, 0.021626237779855728, 0.002192768268287182, -0.001260882243514061, 0.018721995875239372, -0.023493250831961632, 0.030598275363445282, -0.011616972275078297, -0.019331367686390877, 0.01505279541015625, 0.01505279541015625, 0.004524914547801018, 0.0019042887724936008, -0.014962038025259972]
```
parnmatt, line 52: do we need _all_ 1536 dimentions? Can we not have a handful, with some elipses?

The vector we get out of OpenAI will not necessarily be the _exact_ same if we ask again with the _exact_ same input.
One could argue it should be pure semantically, but practically it isn't … but is _very_ close. Similarity is what matters for the vectors rather than exact elementwise equality.
So I'm not necessarily seeing the benefit of rendering the whole thing.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
+If no name is given when created, a random name will be assigned.
+As of Neo4j 5.16, the index name can also be given as a parameter: `CREATE VECTOR INDEX $name ...`
 
+[TIP]
 Creating a vector index requires link:{neo4j-docs-base-uri}/operations-manual/{page-version}/authentication-authorization/database-administration/#access-control-database-administration-index[the `CREATE INDEX` privilege].
 
-[IMPORTANT]
-====
-The index name must be unique among both indexes and constraints.
-====
```
parnmatt: Arguably this should still be noted, just as it is in the create section for the indexes for search performance.
(likewise for fulltext, which I see is missing this note too)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
-[[indexes-vector-create]]
-== Create and configure vector indexes
+A vector index is a single-label, single-property index for nodes or a single-relationship-type, single-property index for relationships.
+It can be used to index nodes by `LIST<INTEGER | FLOAT>` properties valid to the dimensions of the index.
 
-You can create vector indexes using the `CREATE VECTOR INDEX` command.
-An index can be given a unique name when created (or get a generated one), which is used to reference the specific index when xref:#indexes-vector-query[querying] or xref:indexes-vector-drop[dropping] it.
+A vector index is created by using the `CREATE VECTOR INDEX` command.
+It is recommended to give the index a name when it is created.
+If no name is given when created, a random name will be assigned.
```
parnmatt, line 72: however if this is how we're presenting this now, it should be consistent in the indexes for search performance.
Though I think it should still be noted that a chosen name has to be unique among indexes and constraints.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
-The `indexConfig` is a `MAP` from `STRING` values to `STRING` and `INTEGER` values, and is used to set the index-specific configuration settings (the vector dimensions and similarity function).
-
-The command is optionally idempotent. This means that its default behavior is to throw an error if an attempt is made to create the same index twice.
+<1> The `CREATE VECTOR INDEX` command is optionally idempotent.
+This means that its default behavior is to throw an error if an attempt is made to create the same index twice.
 With `IF NOT EXISTS`, no error is thrown and nothing happens should an index with the same name, schema or both already exist.
 It may still throw an error should a constraint with the same name exist.
 As of Neo4j 5.17, an informational notification is instead returned showing the existing index which blocks the creation.
+<2> The `OPTIONS` map is mandatory since a vector index cannot be created without setting the vector dimensions and similarity function.
+In this example, the vector dimension is set to `1536` and the vector similarity function is `cosine`, which generally perfoms best for text embeddings.
```
parnmatt: out of curiosity, where has cosine generally performing best for text embeddings?
What do you mean by performing. Calculation? Accuracy? why text embeddings specifically?
Is this from an article? Could be useful to link to it if so.

cosine and euclidean mean quite different things in terms of similarity … however, when dealing with unit vectors, their order is the same and thus cosine _can_ be faster depending on how it is implemented.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
+A vector index allows you to write queries that match a neighborhood of nodes based on the similarity between the embedding properties of those nodes and the ones specified in the query.
+This similarity is represented by a `score` between `0.0` and `1.0`.
 
-|===
+== Create vector indexes
 
-[[indexes-vector-create]]
-== Create and configure vector indexes
+A vector index is a single-label, single-property index for nodes or a single-relationship-type, single-property index for relationships.
+It can be used to index nodes by `LIST<INTEGER | FLOAT>` properties valid to the dimensions of the index.
```
parnmatt: it's not just valid to the dimensions of the index, but the validity of the vector in regards to the similarity function.

parnmatt: it may not hurt to also have a link / note to the index providers for compatibility section as a reminder that they have different concepts and types that are considered valid.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
 ----
 
-Assume for each review, you have generated a 256-dimensional vector `embedding` of the review's `text` using a shortening of Open AI's `text-embedding-3-large` model.
-This model suggests a xref:indexes-vector-similarity-cosine[cosine similarity].
-For more information, see link:https://platform.openai.com/docs/guides/embeddings/which-distance-function-should-i-use[OpenAI's official documentation].
-
-You can create a cosine relationship vector index over the `embedding` property.
+* The `indexName` refers to the unique name of the vector index to query.
+* The `numberOfNearestNeighbours` refers to the number of nearest neighbors to return as the neighborhood.
+* The `query` vector refers to the `LIST <INTEGER | FLOAT>` in which to search for the neighborhood.
```
parnmatt:
```suggestion
* The `query` vector refers to the `LIST<INTEGER | FLOAT>` in which to search for the neighborhood.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
-CREATE VECTOR INDEX `review-embeddings`
-FOR ()-[r:REVIEWED]-() ON (r.embedding)
-OPTIONS {indexConfig: {
- `vector.dimensions`: 256,
- `vector.similarity_function`: 'cosine'
-}}
+MATCH (m:Movie {title: 'Godfather, The'})
+CALL db.index.vector.queryNodes('moviePlots', 5, m.embedding)
+YIELD node, score
+RETURN node.title AS title, node.plot AS plot, score
```
parnmatt: Using an alias for the returned `node` or `relationship` variable is good practice we want to encourage, especially for readability.
```suggestion
MATCH (m:Movie {title: 'Godfather, The'})
CALL db.index.vector.queryNodes('moviePlots', 5, m.embedding)
YIELD node AS movie, score
RETURN movie.title AS title, movie.plot AS plot, score
```
I've used `movie` here, but could quite easily have `m` be `movie` and use the alias `neighborMovie` or something if you really want, and dislike mixing single letter aliases with word aliases.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
-[[indexes-vector-query]]
-== Query a vector index
+== Show vector indexes
 
-You can query a vector index using the {link-procedures-reference}#procedure_db_index_vector_queryNodes[`db.index.vector.queryNodes`] or the {link-procedures-reference}#procedure_db_index_vector_queryRelationships[`db.index.vector.queryRelationships`] procedure.
+To list all vector indexes in a database, use the `SHOW VECTOR INDEXES` command:
 
-.Signature for `db.index.vector.queryNodes` to query a node vector index
-[source,syntax,role="noheader",indent=0]
+.Show all full-text indexes in a database
```
parnmatt:
```suggestion
.Show all vector indexes in a database
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
+[source,syntax]
+----
+db.index.vector.queryRelationships(indexName :: STRING, numberOfNearestNeighbours :: INTEGER, query :: ANY) :: (relationship :: RELATIONSHIP, score :: FLOAT)
+----
 
-| "review-embeddings" | "VECTOR" | "RELATIONSHIP" | ["REVIEWED"]  | ["embedding"]
-| {indexProvider: "vector-2.0", indexConfig: {vector.dimensions: 256, vector.similarity_function: "cosine"}}
-5+d|Rows: 2
+[TIP]
+To compute the similarity score between two specific vector pairs, without using a vector index, use xref:functions/vector.adoc[].
```
parnmatt: Is this comma needed?
```suggestion
To compute the similarity score between two specific vector pairs without using a vector index, use xref:functions/vector.adoc[].
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
...
 
-|===
+=== Performance suggestions
```
parnmatt: I can understand moving this, but why is this a subsection under Query?
This is a general performance improvement in both index creation _and_ querying.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,302 @@
-:description: Information about how to use vector indexes to perform approximate nearest neighbor search in Neo4j.
+:description: Information about creating, querying, and deleting vector indexes with Cypher.
 :page-role: new-5.13
-:fn-hnsw: footnote:hnsw[http://dx.doi.org/10.1109/TPAMI.2018.2889473[Efficient and Robust Approximate Nearest Neighbor Search Using Hierarchical Navigable Small World Graphs] -- Yury A. Malkov and Dmitry A. Yashunin (preprint: https://arxiv.org/abs/1603.09320:[arXiv:1603.09320])]
-:fn-ieee-754: footnote:ieee-754[https://doi.org/10.1109/IEEESTD.2019.8766229[IEEE Standard for Floating-Point Arithmetic]]
-:fn-provider-differences: footnote:provider-differences[xref:indexes-vector-provider-differences[]]
-:page-role: new-5.13
-:link-operations-manual: link:{neo4j-docs-base-uri}/operations-manual/{page-version}
-:link-procedures-reference: {link-operations-manual}/reference/procedures/
-:ieee-754: IEEE 754{fn-ieee-754}
```
parnmatt, line 9: I see you've removed the linking to IEEE 754, I think it would be useful to link it once at least somewhere, even if it is to the Wikipedia article rather than the standard itself. Perhaps in the similarity section.

parnmatt: I also see you've rendered it as `IEEE-754` rather than `IEEE 754` throughout the document. It should be rendered with a space not a dash as per the IEEE convention, the dash should be used only when specifying a publishing year (if you need to be specific about a revision)
IEEE 754
IEEE 754-2019 (current version)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: We had some trouble with this, and I thought we had fixed it. (I also forgot to change the url to point to the raw file, which is here: https://github.com/neo4j-graph-examples/recommendations/raw/main/data/recommendations-embeddings-50.dump). Will investigate.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: My pov was that it could be interesting for someone newish to the topic, and while it isn't very convenient in the asciidoc file, it renders as one line in the build. I also don't think we claim it would be identical if generated again. In other words, I am inclined to keep it.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: This is what we advice in the GraphAcademy, and I have read the same heuristic elsewhere (e.g. https://www.geeksforgeeks.org/when-to-use-cosine-similarity-over-euclidean-similarity/ - "In summary, use Cosine Similarity when dealing with high-dimensional data, text analysis, or situations where the magnitude of vectors is not crucial", or https://www.baeldung.com/cs/euclidean-distance-vs-cosine-similarity -
" retrieval of the most similar texts to a given document, generally function better with cosine similarity"). 

I am aware that these are not academic references, and I can remove it if you think it is not accurate. My thinking in including it was that it might be a useful guideline for non-experts reading along, who can then at a later stage familiarise themselves with your detailed explanation of the differences between the two, which is still in the document.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: No, it isnt 👍

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: You'll need 5.11+ to use a vector index. I do not mind what version we put up, so long as it's intentional and documented with the correct minimum version. :sweat_smile:

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: That's fair. It is often a safe bet to get something, somewhat useful.

The reason we use cosine with open ai, not only because its what they suggest, but because it's a unit vector (thus magnitude doesn't matter) and that's cosine's bread and butter.

I wouldn't say it's a wrong statement, the main issue I have is to use of "perform" without context, as that means many different things to different people.

Perhaps, something more like: "which is generally the preferred similarity function for text embeddings" to the need for the context, whilst still suggesting it's a good default option to go for.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
stefano-ottolenghi: I've recreated the dump from a 5.13 (the first Neo4j version where vector indexes were GA).

parnmatt: Great! I've confirmed that `data/recommendations-embeddings-50.dump` is 5.13
However, its filename still suggests it's 5.0, it could be argued that the particular file should be named `data/recommendations-embeddings-513.dump` or something like that.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: Yes - I will update the other pages in a separate PR.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: I have added a footnote to the first mention of IEEE 754 in the similarity section

modules/ROOT/images/vector_index_graph.svg
Hunterness, line 1: Mostly a curiosity: It feels a bit odd that you can have `(:Person:Actor)-[:DIRECTED]->(:Movie)` and `(:Person:Director)-[:ACTED_IN]->(:Movie)`, without needing both the `:Actor` and `:Director` labels at the same time 🤔

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
-
+:test-setup-dump: https://github.com/neo4j-graph-examples/recommendations/raw/main/data/recommendations-embeddings-50.dump
 :l2-norm: image:l2.svg["l2"]-norm
 
-[[indexes-vector]]
```
Hunterness, line 13: we don't want/need to keep the tag for reference?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
+To recreate the graph, download and import the link:https://github.com/neo4j-graph-examples/recommendations/blob/main/data/recommendations-embeddings-50.dump[5.0 dump file] to an empty Neo4j database.
+Dump files can be imported for both link:{neo4j-docs-base-uri}/aura/auradb/importing/import-database/[Aura] and link:{neo4j-docs-base-uri}/operations-manual/{page-version}/backup-restore/restore-dump/[on-prem] instances.
 
-| Drop vector index.
-| `+DROP INDEX index_name+`
-| Drop the specified index, see the xref:indexes/search-performance-indexes/managing-indexes.adoc#drop-indexes[`DROP INDEX`] command for details.
+[NOTE]
+The dump file used to load the dataset contains embeddings generated by https://openai.com/[OpenAI], using the model `text-embedding-ada-002`.
+The examples on this page will not show how to generate embeddings, or how to use the procedures `db.create.setNodeVectorProperty` and `db.create.setRelationshipVectorProperty` to set vector properties.
+More information about these topics can be found in *The new DocsAI page for the masses*.
```
Hunterness: `can be found in *The new DocsAI page for the masses*` 🤣

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
-| Set relationship vector property.
-| {link-procedures-reference}#procedure_db_create_setRelationshipVectorProperty[`db.create.setRelationshipVectorProperty`]
-| Update a given relationship property with the given vector in a more space-efficient way than directly using xref:clauses/set.adoc#set-set-a-property[`SET`]. label:beta[] label:new[Introduced in 5.18] 
+Words that are semantically similar or related are often represented by vectors that are closer to each other in this vector space.
+This allows for mathematical operations like addition and subtraction to carry semantic meaning.
+For example, the vector representation of "king" minus "man" plus "woman" might be close to the vector representation of "queen."
+In other words, vector embeddings can be said to be a numerical representation of a particular data object, capturing its semantic meaning.
 
-|===
+A vector index allows you to write queries that match a neighborhood of nodes based on the similarity between the embedding properties of those nodes and the ones specified in the query.
```
Hunterness: should this sentence just talk about nodes or also relationships?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
+This similarity is represented by a `score` between `0.0` and `1.0`.
 
-[[indexes-vector-create]]
-== Create and configure vector indexes
+== Create vector indexes
 
-You can create vector indexes using the `CREATE VECTOR INDEX` command.
-An index can be given a unique name when created (or get a generated one), which is used to reference the specific index when xref:#indexes-vector-query[querying] or xref:indexes-vector-drop[dropping] it.
```
Hunterness, line 98: we no longer want to mention that the name is needed for querying and dropping?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: I think the old text was consistent over both constraint, search, fulltext and vector indexes so just make sure to sync all of them when updating

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
-====
 The index name must be unique among both indexes and constraints.
-====
 
-A vector index is a single-label, single-property index for nodes or a single-relationship-type, single-property index for relationships.{fn-provider-differences}
-A vector index needs to be configured with both the dimensionality of the vector (`INTEGER` between `1` and `4096` _inclusive_),{fn-provider-differences} and the measure of similarity between two vectors (case-insensitive `STRING`).
-For details, see xref:#indexes-vector-similarity[].
-
-
-.Syntax for creating vector indexes
```
Hunterness, line 112: have you removed the syntax descriptions competly or has it moved? Cause I don't think we should remove them fully but instead have them somewhere at least for quick reference

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
-For more information, see link:https://platform.openai.com/docs/guides/embeddings/which-distance-function-should-i-use[OpenAI's official documentation].
-
-You can create a cosine node vector index over the `embedding` property.
-
-.Query
-[source,cypher]
-----
-CREATE VECTOR INDEX `abstract-embeddings`
-FOR (n: Abstract) ON (n.embedding)
+FOR ()-[r:REL_TYPE]-() ON (r.embedding)
```
Hunterness, line 114: I think you're missing the 
```
CREATE VECTOR INDEX name IF NOT EXISTS
```
part of this query

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
 With `IF NOT EXISTS`, no error is thrown and nothing happens should an index with the same name, schema or both already exist.
 It may still throw an error should a constraint with the same name exist.
 As of Neo4j 5.17, an informational notification is instead returned showing the existing index which blocks the creation.
+<2> The `OPTIONS` map is mandatory since a vector index cannot be created without setting the vector dimensions and similarity function.
+In this example, the vector dimension is set to `1536` and the vector similarity function is `cosine`, which is generally the preferred similarity function for text embeddings.
+To read more about the available similarity functions, see xref:indexes/semantic-indexes/vector-indexes.adoc#similarity-functions[].
 
 [NOTE]
-====
-The new index is not immediately available but is created in the background.
```
Hunterness, line 160: We don't want to keep this note around?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
-The measure of similarity is determined by the given vector similarity function.
-This defines how similar two vectors are to one another by a similarity score, how vectors are interpreted, and what vectors are valid for the index.
-
-A node or relationship is indexed if all the following are true:
-
-* The node/relationship contains the configured label/relationship type.
-* The node/relationship contains the configured property key.
-* The respective property value is of type `LIST<INTEGER | FLOAT>`.{fn-provider-differences}
-* The xref:functions/scalar.adoc#functions-size[`size()`] of the respective value is the same as the configured dimensionality.
-* The value is a valid vector for the configured similarity function.
```
Hunterness, line 173: or keep this information about what get indexed?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -554,61 +359,50 @@ Cosine similarity rebounds the trigonometric cosine.
...
-
-| xref:indexes-vector-similarity-cosine[Cosine similarity vector validity]
-| All vector components can be represented finitely in {ieee-754} *_single_* precision.
-
-{nbsp}
-
-Its {l2-norm} is non-zero and can be represented finitely in {ieee-754} *_single_* precision.
+| Create node vector index.
+| link:{neo4j-docs-base-uri}/operations-manual/{page-version}/reference/procedures/#procedure_db_index_vector_createnodeindex[`db.index.vector.createNodeIndex`]
+| Create a vector index for the specified label and property with the given vector dimensionality using the given similarity function.
```
Hunterness, line 396: we don't want to mention for the ones that have been replaced that they are replaced?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
-====
+.Configuration settings
+[source,config]
+----
+server.jvm.additional=--add-modules jdk.incubator.vector
+----
 
-[[indexes-vector-query]]
-== Query a vector index
+== Show vector indexes
```
Hunterness, line 192: we should mention the privilege needed for show indexes, and it might also be nice to mention that it is the same show command as the general one, just with the vector type filtering

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
-| Sarah Lee | 0.8123456835746765
+For a full description of all return columns, see xref:indexes/search-performance-indexes/managing-indexes.adoc#listing-indexes-result-columns[Search-performance indexes -> Result columns for listing indexes].
 
-| David Wilson | 0.7865431904792786
 
-| Jennifer Davis | 0.7690123319625854
+== Drop vector indexes
 
-| Christopher Evans | 0.7356789112091064
+A vector index is dropped by using the xref:indexes/search-performance-indexes/managing-indexes.adoc#drop-an-index[same command as for other indexes], `DROP INDEX`.
```
Hunterness, line 258: I guess that since we say it's the same command we don't need to mention the needed privileges here, but since we have the sentence about parameter names maybe it might be nice?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
-(:Manager)-[:REVIEWED]->(:Employee)
+db.index.vector.queryNodes(indexName :: STRING, numberOfNearestNeighbours :: INTEGER, query :: ANY) :: (node :: NODE, score :: FLOAT)
 ----
 
-Assume for each review, you have generated a 256-dimensional vector `embedding` of the review's `text` using a shortening of Open AI's `text-embedding-3-large` model.
-This model suggests a xref:indexes-vector-similarity-cosine[cosine similarity].
-For more information, see link:https://platform.openai.com/docs/guides/embeddings/which-distance-function-should-i-use[OpenAI's official documentation].
+* The `indexName` refers to the unique name of the vector index to query.
+* The `numberOfNearestNeighbours` refers to the number of nearest neighbors to return as the neighborhood.
+* The `query` vector refers to the `LIST<INTEGER | FLOAT>` in which to search for the neighborhood.
```
Hunterness, line 134: feels odd to only have these kinds of descriptions for the node procedure signature, but not at all for the relationship procedure signature

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,307 @@
...
+| 2  | "moviePlots" | "ONLINE" | 100.0             | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"] | "vector-2.0"  | NULL             | 2024-05-07T09:19:09.225Z | 47        |
++----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 ----
 
-* The `indexName` (a `STRING`) refers to the unique name of the vector index to query.
-* The `numberOfNearestNeighbours` (an `INTEGER`) refers to the number of nearest neighbors to return as the neighborhood.
-* The `query` vector (a `LIST<INTEGER | FLOAT>`) in which to search for the neighborhood.
-
-The procedures return the neighborhood of nodes or relationships with their respective similarity scores, ordered by those scores.
-The scores are bounded between `0` and `1`, where the closer to `1` the score is, the more similar the indexed vector is to the query vector.
```
Hunterness, line 287: we don't want to define the return values of the procedures anymore?

Hunterness (commented):
I tried to compare each section by itself with the current docs (as the git changes here is a bit of a mess and hard to get an overview of what is actually changed for each section)

I still want to do an overall look through the new file to see how it fits within itself but that'll have to wait for later (hopefully tomorrow). But these are the comments I have so far on the comparisons (no guarantee I haven't missed some parts to compare XD)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: Haha yeah - will change when said page for the masses is available, which is soon 😄

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: My thought was that this is obvious to anyone reading the relevant sections for query, show, and drop.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: No, the syntax is here: https://neo4j.com/docs/cypher-manual/current/indexes/syntax/#create-vector-index

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: I think we should be consise, and this information can be inferred elsewhere on the page.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: It would just be a repetition of information, though? I think having it once for the node procedure is sufficient.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: That should probably be there - will re-add.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: before when we had the signatures beside each other I definitely agree, maybe have a one line about it being the same ones for the relationship as node down by the relationship signature?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: Fair, it can be just over there, just didn't want it to be completely gone 👍

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: I think we say something like it in the other index/constraints sections but maybe that has been removed at some point as well 🤷

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: I guessed something along those lines XD

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: (goes for all that were removed)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -178,6 +183,9 @@ server.jvm.additional=--add-modules jdk.incubator.vector
 
 To list all vector indexes in a database, use the `SHOW VECTOR INDEXES` command:
```
Hunterness:
```suggestion
To list all vector indexes in a database, use the `SHOW VECTOR INDEXES` command.
```
or maybe something like
```suggestion
To list all vector indexes in a database, use the `SHOW VECTOR INDEXES` command.
This is xref:indexes/search-performance-indexes/managing-indexes.adoc#list-indexes[the same `SHOW` command as for other indexes] with the index type filtering on `VECTOR`.
```
with a sentence of some kind referencing the regular show command

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
-:ieee-754: IEEE 754{fn-ieee-754}
-
+:test-setup-dump: https://github.com/neo4j-graph-examples/recommendations/raw/main/data/recommendations-embeddings-50.dump
 :l2-norm: image:l2.svg["l2"]-norm
 
-[[indexes-vector]]
-= Vector search indexes
+= Vector indexes
 
 _Node vector search indexes were released as a public beta in Neo4j 5.11 and general availability in Neo4j 5.13._
```
Hunterness, line 8: should we mention when relationship vector indexes were introduced here as well?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
+A vector index allows you to write queries that match a neighborhood of nodes or relationships based on the similarity between the embedding properties of those nodes or relationships and the ones specified in the query.
 
-[[indexes-vector-create]]
-== Create and configure vector indexes
+== Create vector indexes
 
-You can create vector indexes using the `CREATE VECTOR INDEX` command.
-An index can be given a unique name when created (or get a generated one), which is used to reference the specific index when xref:#indexes-vector-query[querying] or xref:indexes-vector-drop[dropping] it.
+A vector index is a single-label, single-property index for nodes or a single-relationship-type, single-property index for relationships.
+It can be used to index nodes by `LIST<INTEGER | FLOAT>` properties valid to the dimensions of the index and the vector similarity function.
```
Hunterness:
```suggestion
It can be used to index nodes by `LIST<INTEGER | FLOAT>` properties valid to the dimensions and vector similarity function of the index.
```
I feel like that reads better?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
-ON (n.propertyName)
-OPTIONS "{" option: value[, ...] "}"
-----
-| Create a vector index on nodes.
-The options map is mandatory because setting the vector dimensions and similarity function is mandatory when creating a vector index.
+[TIP]
+Creating indexes requires link:{neo4j-docs-base-uri}/operations-manual/{page-version}/authentication-authorization/database-administration/#access-control-database-administration-index[the `CREATE INDEX` privilege].
 
-| [source, syntax, role=noplay, indent=0]
+.Create vector index for `Movie` nodes on the `embedding` property label:new[Introduced in 5.15]
```
Hunterness, line 81: it feels a bit odd to first read that they are available for general availability in 5.13 and then see 5.15 label on the example with no comment on the command being introduced in 5.15 prior to that

Hunterness: maybe just add something to the ```A vector index is created by using the `CREATE VECTOR INDEX` command.``` sentence about the command being introduced in 5.15? (and that it was created through a procedure before that?)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
-The `CREATE VECTOR INDEX` command take an `OPTIONS` clause. This has two parts, the `indexProvider` and `indexConfig`.
-As of Neo4j 5.18, there are two available index providers, `vector-2.0` (default) and `vector-1.0`.
-The `indexConfig` is a `MAP` from `STRING` values to `STRING` and `INTEGER` values, and is used to set the index-specific configuration settings (the vector dimensions and similarity function).
-
-The command is optionally idempotent. This means that its default behavior is to throw an error if an attempt is made to create the same index twice.
+<1> The `CREATE VECTOR INDEX` command is optionally idempotent.
+This means that its default behavior is to throw an error if an attempt is made to create the same index twice.
 With `IF NOT EXISTS`, no error is thrown and nothing happens should an index with the same name, schema or both already exist.
 It may still throw an error should a constraint with the same name exist.
 As of Neo4j 5.17, an informational notification is instead returned showing the existing index which blocks the creation.
```
Hunterness: I know you didn't change this bit but as I was reading over it I did find it sounding a bit odd. Maybe either rewriting the last sentence or switch order of the sentences helps?

```suggestion
With `IF NOT EXISTS`, no error is thrown and nothing happens should an index with the same name, schema or both already exist.
It may still throw an error should a constraint with the same name exist.
As of Neo4j 5.17, an informational notification is returned in the case when nothing happens, showing the existing index which blocks the creation.
```

```suggestion
With `IF NOT EXISTS`, no error is thrown and nothing happens should an index with the same name, schema or both already exist.
As of Neo4j 5.17, an informational notification is instead returned showing the existing index which blocks the creation.
It may still throw an error should a constraint with the same name exist.
```

(this update might need to be synced across constraints and remaining indexes as well when you sync the other parts)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: found the note further down 🤔 might still be nicer to mention something about it earlier?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
+OPTIONS {indexConfig: {
+ `vector.dimensions`: $dimension,
+ `vector.similarity_function`: $similarityFunction
+}}
 ----
 
-Assume for each abstract, you have generated a 1536-dimensional vector `embedding` of the abstract's `text` using Open AI's `text-embedding-ada-002` model.
-This model suggests a xref:indexes-vector-similarity-cosine[cosine similarity].
-For more information, see link:https://platform.openai.com/docs/guides/embeddings/which-distance-function-should-i-use[OpenAI's official documentation].
+A newly created index is not immediately available but is created in the background.
```
Hunterness: I would put this comment up by the node parts since all other info is up there and then the relationship one just shows the syntax

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
+* The `query` vector refers to the `LIST<INTEGER | FLOAT>` in which to search for the neighborhood.
 
-Assume you have a graph of employees and their managers, who are themselves employees.
-Managers review their reports periodically, and you wish to search for reviews of similar themes and nuances to find excellent employees.
+The procedure returns the neighborhood of nodes with their respective similarity scores, ordered by those scores.
+The scores are bounded between `0` and `1`, where the closer to `1` the score is, the more similar the indexed vector is to the query vector.
 
-.Data model
-[source,syntax,role="noheader",indent=0]
+.Find the 5 movies with the most similar plot to The Godfather
```
Hunterness, line 139: Feels a bit weird to have no text above saying `here's an example` or something marking it as an example 🤷 but I guess we have in general gotten less and less of that as the docs has been restructured/re-formatted. I think previously examples had boxes around them and such so might just be leftover feelings from the more structured versions :P

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
+A vector index is a single-label, single-property index for nodes or a single-relationship-type, single-property index for relationships.
+It can be used to index nodes by `LIST<INTEGER | FLOAT>` properties valid to the dimensions of the index and the vector similarity function.
+Note that the available vector index providers support different index schemas, property value types, and vector dimensionalities.
+For more information, see xref:indexes/semantic-indexes/vector-indexes.adoc#_vector_index_providers_for_compatibility[].
 
-Creating a vector index requires link:{neo4j-docs-base-uri}/operations-manual/{page-version}/authentication-authorization/database-administration/#access-control-database-administration-index[the `CREATE INDEX` privilege].
+A vector index is created by using the `CREATE VECTOR INDEX` command.
+It is recommended to give the index a name when it is created.
+If no name is given when created, a random name will be assigned.
+As of Neo4j 5.16, the index name can also be given as a parameter: `CREATE VECTOR INDEX $name ...`
```
Hunterness:
```suggestion
As of Neo4j 5.16, the index name can also be given as a parameter: `CREATE VECTOR INDEX $name ...`.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
-[source,syntax,role="noheader",indent=0]
+.Result
+[role=queryresult]
 ----
-db.index.vector.queryRelationships(indexName :: STRING, numberOfNearestNeighbours :: INTEGER, query :: ANY) :: (relationship :: RELATIONSHIP, score :: FLOAT)
++----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| id | name         | state    | populationPercent | type     | entityType | labelsOrTypes | properties        | indexProvider | owningConstraint | lastRead                 | readCount |
++----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 2  | "moviePlots" | "ONLINE" | 100.0             | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"] | "vector-2.0"  | NULL             | 2024-05-07T09:19:09.225Z | 47        |
++----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
Hunterness:
```suggestion
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id | name         | state    | populationPercent | type     | entityType | labelsOrTypes | properties        | indexProvider | owningConstraint | lastRead                 | readCount |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 2  | "moviePlots" | "ONLINE" | 100.0             | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"]     | "vector-2.0"  | NULL             | 2024-05-07T09:19:09.225Z | 47        |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
or 
```suggestion
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id | name         | state    | populationPercent | type     | entityType | labelsOrTypes | properties    | indexProvider | owningConstraint | lastRead                 | readCount |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 2  | "moviePlots" | "ONLINE" | 100.0             | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"] | "vector-2.0"  | NULL             | 2024-05-07T09:19:09.225Z | 47        |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

Hunterness: to not get the column sizes out of order
image

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
+| 2  | "moviePlots" | "ONLINE" | 100.0             | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"] | "vector-2.0"  | NULL             | 2024-05-07T09:19:09.225Z | 47        | 2024-05-07T08:26:19.072Z | {indexConfig: {`vector.dimensions`: 1536, `vector.similarity_function`: "COSINE"}, indexProvider: "vector-2.0"} | ""             | "CREATE VECTOR INDEX `moviePlots` FOR (n:`Movie`) ON (n.`embedding`) OPTIONS {indexConfig: {`vector.dimensions`: 1536,`vector.similarity_function`: 'COSINE'}, indexProvider: 'vector-2.0'}" |
++---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
Hunterness: same here
```suggestion
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id | name         | state    | populationPercent | type     | entityType | labelsOrTypes | properties        | indexProvider | owningConstraint | lastRead                 | readCount | trackedSince             | options                                                                                                         | failureMessage | createStatement                                                                                                                                                                                  |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 2  | "moviePlots" | "ONLINE" | 100.0             | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"]     | "vector-2.0"  | NULL             | 2024-05-07T09:19:09.225Z | 47        | 2024-05-07T08:26:19.072Z | {indexConfig: {`vector.dimensions`: 1536, `vector.similarity_function`: "COSINE"}, indexProvider: "vector-2.0"} | ""             | "CREATE VECTOR INDEX `moviePlots` FOR (n:`Movie`) ON (n.`embedding`) OPTIONS {indexConfig: {`vector.dimensions`: 1536,`vector.similarity_function`: 'COSINE'}, indexProvider: 'vector-2.0'}"     |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
or
```suggestion
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id | name         | state    | populationPercent | type     | entityType | labelsOrTypes | properties    | indexProvider | owningConstraint | lastRead                 | readCount | trackedSince             | options                                                                                                         | failureMessage | createStatement                                                                                                                                                                              |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 2  | "moviePlots" | "ONLINE" | 100.0             | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"] | "vector-2.0"  | NULL             | 2024-05-07T09:19:09.225Z | 47        | 2024-05-07T08:26:19.072Z | {indexConfig: {`vector.dimensions`: 1536, `vector.similarity_function`: "COSINE"}, indexProvider: "vector-2.0"} | ""             | "CREATE VECTOR INDEX `moviePlots` FOR (n:`Movie`) ON (n.`embedding`) OPTIONS {indexConfig: {`vector.dimensions`: 1536,`vector.similarity_function`: 'COSINE'}, indexProvider: 'vector-2.0'}" |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
-| Emily Johnson | 0.8456789255142212
-
-| Michael Brown | 0.8245679140090942
+[role=queryresult]
+----
++--------------------------------------------------------------------------+
+| name         | type     | entityType | labelsOrTypes | properties        |
++--------------------------------------------------------------------------+
+| "moviePlots" | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"] |
++--------------------------------------------------------------------------+
```
Hunterness: and here
```suggestion
+--------------------------------------------------------------------------+
| name         | type     | entityType | labelsOrTypes | properties        |
+--------------------------------------------------------------------------+
| "moviePlots" | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"]     |
+--------------------------------------------------------------------------+
```
or
```suggestion
+----------------------------------------------------------------------+
| name         | type     | entityType | labelsOrTypes | properties    |
+----------------------------------------------------------------------+
| "moviePlots" | "VECTOR" | "NODE"     | ["Movie"]     | ["embedding"] |
+----------------------------------------------------------------------+
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
-DROP INDEX `abstract-embeddings`
-----
+{nbsp}
 
-.Result
-[queryresult]
-----
-Removed 1 index.
-----
+_No relationship support_
```
Hunterness:
```suggestion
_No relationship support._
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
+== Vector index providers for compatibility
 
-2+d| Rows: 10
+As of Neo4j 5.18, the default and preferred vector index provider is `vector-2.0`.
+Previously created `vector-1.0` indexes will continue to function.
+New indexes can still be created with the `vector-1.0` provider if it is specified in the `OPTIONS` map.
 
-|===
+.Learn more about vector index provider differences
+[%collapsible]
```
Hunterness, line 278: I like the collapsible things :D it's a nice function :)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: though I guess it is technically only the dimension of the index that will ensure the node get indexed

Hunterness:
```suggestion
It can be used to index nodes or relationships by `LIST<INTEGER | FLOAT>` properties valid to the dimensions of the index and the vector similarity function.
```
in addition to my comment on the second half of the sentence, maybe it should also mention both nodes and relationships as per the sentence above it

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -1,503 +1,317 @@
...
+| Indexed vector dimensionality
+| `INTEGER` between `1` and `2048` _inclusive_.
+| `INTEGER` between `1` and `4096` _inclusive_.
 
-.Signature for `db.create.setVectorProperty` label:deprecated[]
-[source,syntax]
-----
-db.create.setVectorProperty(node :: NODE, key :: STRING, vector :: ANY) :: (node :: NODE)
-----
+| xref:indexes-vector-similarity-cosine[Cosine similarity vector validity]
```
Hunterness: I don't think this link resolves anymore

Hunterness: I have not checked all links so there might be others that no longer works as well 🤷

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -633,9 +438,7 @@ The following table lists the known issues and the version in which they were fi
 
 | Procedure signatures from `SHOW PROCEDURES` will render the vector arguments with a type of `ANY` rather than the semantically correct type of `LIST<INTEGER \| FLOAT>`.
 [NOTE]
-====
 The types are still enforced as `LIST<INTEGER \| FLOAT>`.
-====
 |
```
Hunterness: I can see that this is left from before, but I think this table should only show fixed issues and when they were fixed, unfixed issues were up in the limitations list before

and I assume since this doesn't have a `Fixed in` version that it is not yet fixed, so either it should move up or the sentence above the column should mention something about `and if fixed, the version in which they were fixed:` instead of just `and the version in which they were fixed:` if not everything in the table has been fixed

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -648,36 +451,31 @@ The types are still enforced as `LIST<INTEGER \| FLOAT>`.
...
 
 [TIP]
-====
 Use the procedure `db.index.vector.createNodeIndex` to create the a vector index.
 Procedure signature:
 [source,syntax,role="noheader"]
 ----
 db.index.vector.createNodeIndex(indexName :: STRING, label :: STRING, propertyKey :: STRING, vectorDimension :: INTEGER, vectorSimilarityFunction :: STRING)
 ----
-====
```
Hunterness: without this grouping it looks very odd
image
vs the old
image
which clearly showed the signature belonging to the tip and not freestanding after (especially since the end of the ip says `Procedure signature:`)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -648,36 +451,31 @@ The types are still enforced as `LIST<INTEGER \| FLOAT>`.
...
 
 [TIP]
-====
 Filtering on vector indexes can be done with the `WHERE` clause as well:
 [source,cypher]
 ----
 SHOW INDEXES
 WHERE type = 'VECTOR'
 ----
-====
```
Hunterness, line 671: same here for the show index one

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -742,7 +532,6 @@ java.lang.IllegalArgumentException: Did not recognize index type VECTOR
 ----
 
 [TIP]
-====
 If a store copy is required, make a note of the information in the `createStatement` column returned from the `SHOW INDEX` command.
 For example:
 [source,cypher]
```
Hunterness: same here as above, without the grouping the example isn't in the tip box which feels wrong as it is part of the tip

Hunterness (commented):
Looked over the built docs page for the overall view and had a few more comments XD

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: It is both the dimension and what the vector similarity function considers valid; for example the zero vector is valid for euclidean, but invalid for cosine.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: indeed, this was the original intent of the code block with the tip, and why we wrote it as such.

JPryce-Aklundh: This is my mistake - will fix.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: I don't think that is necessary - the labelling below makes that clear. This is only to distinguish between beta and GA.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: I think it suffices as it stands - In general I am wary of frontloading information.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: Yes, I think it is pretty clear that it is an example without the example box. The example boxes are more called for when an example requires multiple consecutive queries.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: Good catch!

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: No - I recently learnt these are not necessary (just add a _ in front of the header-title instead).

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: Yes, at least for indexes

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: what readers will get from knowing that it is the same command - they are interested in showing a vector index here, the awareness of it being a similar command to show, e.g, a point index seems redundant.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: I think it's more todo with the view of it that we have (at least previously) tried to put out: the type filtering is just syntax sugar for `WHERE type = 'VECTOR'` and not actually a  separate command. so mentioning that this the same command with filtering on vector indexes keeps that view up instead of "helping" users see them as two separate commands (and start wanting special handling for each index type)

which is also why we have the same set of columns regardless of which type you filter on (and not for example adding columns for dimension or similarity function to `SHOW VECTOR INDEXES` as it is just the `SHOW INDEXES` command with filtering, we did recently get some slack questions around this)

Hunterness: So it's not that there is a similar command, but that it is just part of the same command 🤷

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -505,18 +530,22 @@ The procedures are still usable, just not visible.
...
+| link:{neo4j-docs-base-uri}/operations-manual/{page-version}/#procedure_db_index_vector_queryNodes[`db.index.vector.queryNodes`] query vector validation is incorrect with a xref:indexes/semantic-indexes/vector-indexes.adoc#similarity-functions[cosine] vector index. The {l2-norm} validation only considers the last component of the vector. If that component is `&plusmn;0.0`, an otherwise valid query vector will be thrown as invalid. This can also result in some invalid vectors being used to query, and return a similarity score of `&plusmn;0.0`.
 
 [TIP]
+--
 For {l2-norm}alized vectors (unit vectors), thus having unit length image:l2norm_is_1.svg["The l2-norm of vector v equals 1"], Euclidean and cosine similarity functions produce the same similarity ordering.
-It is _recommended_ to normalize your vectors (if needed), and use a xref:indexes-vector-similarity-euclidean[Euclidean] vector index.
+It is _recommended_ to normalize your vectors (if needed), and use a  xref:indexes/semantic-indexes/vector-indexes.adoc#similarity-functions[Euclidean] vector index.
```
Hunterness:
```suggestion
It is _recommended_ to normalize your vectors (if needed), and use a xref:indexes/semantic-indexes/vector-indexes.adoc#similarity-functions[Euclidean] vector index.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: I think one thing that might be less nice without the box is that we also don't have the `Example X: ...` counter so you can refer to the examples on a page by number, so now you'd have to refer to them by description or query or something else 🤷 

Not that much of an issue, just a reflection 🤷

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
Hunterness: I saw you added back a lot of them but not this one that the comment was on?

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

Hunterness (approved):

parnmatt (approved):

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-954.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-954.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:

PR #1009, opened 2024-07-29
Add new vector index settings

parnmatt:
We have exposed quantization, and `M` and `efConstruction` hyperparameters for the vector index.
We have also allowed the similarity function to be defaulted to `'cosine'`, and for the dimensions to not need to be specified.


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

parnmatt:
I have no strong preference on the wording currently, just jotted down some things we can iterated upon.
I have used description lists for the index settings for now; however, I am open to a different format if preferred.
I am unsure if we want it to be in a collapsible section (or perhaps just the advanced) or if all settings should be clearly visible.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -103,8 +103,10 @@ This means that its default behavior is to throw an error if an attempt is made
 With `IF NOT EXISTS`, no error is thrown and nothing happens should an index with the same name, schema or both already exist.
 It may still throw an error should a constraint with the same name exist.
 As of Neo4j 5.17, an informational notification is returned when nothing happens, showing the existing index which blocks the creation.
-<2> The `OPTIONS` map is mandatory since a vector index cannot be created without setting the vector dimensions and similarity function.
-In this example, the vector dimension is set to `1536` and the vector similarity function is `cosine`, which is generally the preferred similarity function for text embeddings.
+<2> Prior to Neo4j 5.23, the `OPTIONS` map was mandatory since a vector index could not be created without setting the vector dimensions and similarity function.
+Since Neo4j 5.23, the dimensions can be omitted, and the similarity function will default to `'cosine'` if not specified.
```
neo-ionut: Should we say what happens with dimensions if it omitted, as we do for similarity function?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,56 @@ You can also create a vector index for relationships with a particular type on a
...
 ----
 
+[[configuration-settings]]
+=== Configuration settings
+
+`vector.dimensions` ::
+The dimensionality of the vectors to be indexed, see xref:indexes/semantic-indexes/vector-indexes.adoc#embeddings[].
+Setting this value will ensure only vectors with the configured dimensionality will be indexed, and querying the index with a vector of a different dimensionality will return an error. It is recommended to provide the dimensions.
+Accepted values::: `INTEGER` between `1` and `4096` inclusively.
+Default Value::: None, though the setting can be omitted. The setting was mandatory prior to Neo4j 5.23.
```
neo-ionut: It's not clear what happens if it's omitted from this text.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,56 @@ You can also create a vector index for relationships with a particular type on a
...
+`vector.similarity_function` ::
+The name of the similarity function to use to define how similar two vectors are.
+To read more about the available similarity functions, see xref:indexes/semantic-indexes/vector-indexes.adoc#similarity-functions[].
+
+Accepted values::: `STRING`: `'cosine'`, `'euclidean'`.
+Default Value::: `'cosine'`. The setting was mandatory prior to Neo4j 5.23.
+
+`vector.quantization.enabled` label:new[Introduced in 5.23] ::
+Quantization is a technique to reduce the size of vector representations.
+Enabling quantization can accelerate search performance but can slightly decrease accuracy. On machines with limited memory, this is a recommended to always be enabled.
```
neo-ionut: typo

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,56 @@ You can also create a vector index for relationships with a particular type on a
...
+
+`vector.hnsw.ef_construction` label:new[Introduced in 5.23] ::
+The number of nearest neighbors tracked during the insertion of vectors into the HNSW graph.
+Increasing this value increases the quality of the index, and may lead to greater accuracy (with diminishing returns) at the expense of increased index population and update times.
+Vector indexes created prior to Neo4j 5.23 have this setting effectively set as `100`.
+
+Accepted values::: `INTEGER` between `1` and `3200` inclusively.
+Default value::: `100`
+
+For differences between `vector.dimensions` and `vector.similarity_function` between vector index providers, see xref:indexes/semantic-indexes/vector-indexes.adoc#vector-index-providers[].
```
neo-ionut: Should this be here?

neo-ionut (approved):

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: moving a more generic version to the top of the section.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: I think we don't want to focus too much here. Instead I've drawn the focus to the config section where that information is encoded.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,57 @@ You can also create a vector index for relationships with a particular type on a
...
 
+[[configuration-settings]]
+=== Configuration settings
+
+For differences with accepted values between vector index providers, see xref:indexes/semantic-indexes/vector-indexes.adoc#vector-index-providers[].
+
+`vector.dimensions` ::
+The dimensionality of the vectors to be indexed, see xref:indexes/semantic-indexes/vector-indexes.adoc#embeddings[].
+Omitting this setting is possible, and any `LIST<INTEGER | FLOAT>` can be indexed and queried, separated by their dimensionality.
+Setting this value will add additional checks that ensure only vectors with the configured dimensionality will be indexed, and querying the index with a vector of a different dimensionality will return an error. It is recommended to provide the dimensions.
```
rsill-neo4j:
```suggestion
Setting this value adds additional checks that ensure only vectors with the configured dimensionality are indexed, and querying the index with a vector of a different dimensionality returns an error. It is recommended to provide the dimensions.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,57 @@ You can also create a vector index for relationships with a particular type on a
...
+Accepted values::: `INTEGER` between `1` and `4096` inclusively.
+Default Value::: None, though the setting can be omitted. The setting was mandatory prior to Neo4j 5.23.
```
rsill-neo4j:
```suggestion
Default value::: None, though the setting can be omitted. The setting was mandatory prior to Neo4j 5.23.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,57 @@ You can also create a vector index for relationships with a particular type on a
...
+
+`vector.similarity_function` ::
+The name of the similarity function to use to define how similar two vectors are.
```
rsill-neo4j:
```suggestion
The name of the similarity function to use to assess how similar two vectors are.
```

open for discussion
was thinking about "measure", but after a bit of googling wording, i went with "assess", based on this:

https://kdb.ai/learning-hub/articles/methods-of-similarity/

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,57 @@ You can also create a vector index for relationships with a particular type on a
...
+To read more about the available similarity functions, see xref:indexes/semantic-indexes/vector-indexes.adoc#similarity-functions[].
+
+Accepted values::: `STRING`: `'cosine'`, `'euclidean'`.
+Default Value::: `'cosine'`. The setting was mandatory prior to Neo4j 5.23.
```
rsill-neo4j:
```suggestion
Default value::: `'cosine'`. The setting was mandatory prior to Neo4j 5.23.
```

prior to 5.23, there was no default value?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,57 @@ You can also create a vector index for relationships with a particular type on a
...
+
+`vector.quantization.enabled` label:new[Introduced in 5.23] ::
+Quantization is a technique to reduce the size of vector representations.
+Enabling quantization can accelerate search performance but can slightly decrease accuracy. This is recommended to always be enabled on machines with limited memory.
```
rsill-neo4j:
```suggestion
Enabling quantization can accelerate search performance but can slightly decrease accuracy.
This is recommended to always be enabled on machines with limited memory.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -463,18 +508,32 @@ The following table lists the known issues and, if fixed, the version in which t
 |===
 | Known issues | Fixed in
 
+| Using the legacy procedure link:{neo4j-docs-base-uri}/operations-manual/{page-version}/reference/procedures/#procedure_db_index_vector_createnodeindex[`db.index.vector.createNodeIndex`] on a database created prior to Neo4j 5.11 in Neo4j 5.18 and later, before the database has been upgraded, can cause an error.
+In Neo4j 5.20, the error was clarified.
+[TIP]
+--
+Using the `CREATE VECTOR INDEX` command instead will avoid this issue, and would also trigger the upgrade.
```
rsill-neo4j:
```suggestion
Using the `CREATE VECTOR INDEX` command instead avoids this issue and also triggers the upgrade.
```
double-check if the intended meaning is preserved

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -463,18 +508,32 @@ The following table lists the known issues and, if fixed, the version in which t
...
+
+If the use of the procedure is out of the user's control, we need to trigger an upgrade.
+The simplest would be to write to the database on the newer binary.
```
rsill-neo4j:
```suggestion
The simplest is to write to the database on the newer binary.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -463,18 +508,32 @@ The following table lists the known issues and, if fixed, the version in which t
...
+An alternative would be to use link:{neo4j-docs-base-uri}/operations-manual/{page-version}/tools/neo4j-admin/migrate-database/[`neo4j-admin database migrate`] if you have direct access to the machine running the Neo4j DBMS.
```
rsill-neo4j:
```suggestion
An alternative is to use link:{neo4j-docs-base-uri}/operations-manual/{page-version}/tools/neo4j-admin/migrate-database/[`neo4j-admin database migrate`] if you have direct access to the machine running the Neo4j DBMS.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -463,18 +508,32 @@ The following table lists the known issues and, if fixed, the version in which t
 |===
 | Known issues | Fixed in
 
+| Using the legacy procedure link:{neo4j-docs-base-uri}/operations-manual/{page-version}/reference/procedures/#procedure_db_index_vector_createnodeindex[`db.index.vector.createNodeIndex`] on a database created prior to Neo4j 5.11 in Neo4j 5.18 and later, before the database has been upgraded, can cause an error.
+In Neo4j 5.20, the error was clarified.
+[TIP]
+--
+Using the `CREATE VECTOR INDEX` command instead will avoid this issue, and would also trigger the upgrade.
+
+If the use of the procedure is out of the user's control, we need to trigger an upgrade.
```
rsill-neo4j:
```suggestion
If the use of the procedure is out of the user's control, trigger an upgrade.
```

line 519 uses "you", so i went with imperative to omit the "we"

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: correct, both dimensions and similarity function were mandatory to be explicitly provided: there were no defaults.

The index needed to know the dimensionality, and the similarity function was something the user _should_ really explicitly choose; as they had to specify one, it made sense to force the other … as it wasn't clear what the default should be. Hense the original choice.

We made changes in `vector-2.0` that allowed us to potentially support having the dimensions be optionally specified.
As we were making the new things defaultable, and had to massively refactor validation, we decided to act on the potential plan to make similarity defaultable (now there was a clear market default), and dimensionality optional.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: I am unsure if we use define elsewhere. In a way all are correct.

A similarity function assesses similarity, and we measure the similarity between two vectors using a similarity function. Though there are multiple similarity functions, each have different definitions and assess similarity differently, and many are valid (it depends on the question you're asking).
We _define_ what "similar" means for us.

However, I am happy with both alternatives. I think "assess" fits best personally.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,58 @@ You can also create a vector index for relationships with a particular type on a
...
+OPTIONS { indexConfig: {
  `vector.dimensions`: $dimension,
  `vector.similarity_function`: $similarityFunction
 }}
 ----
 
+[[configuration-settings]]
+=== Configuration settings
+
+For differences with accepted values between vector index providers, see xref:indexes/semantic-indexes/vector-indexes.adoc#vector-index-providers[].
```
JPryce-Aklundh: I found this sentence a bit difficult to follow.
```suggestion
For more information about the values accepted by different index providers, see xref:indexes/semantic-indexes/vector-indexes.adoc#vector-index-providers[].
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,58 @@ You can also create a vector index for relationships with a particular type on a
...
+
+`vector.dimensions` ::
+The dimensionality of the vectors to be indexed, see xref:indexes/semantic-indexes/vector-indexes.adoc#embeddings[].
+Omitting this setting is possible, and any `LIST<INTEGER | FLOAT>` can be indexed and queried, separated by their dimensionality.
+Setting this value adds additional checks that ensure only vectors with the configured dimensionality are indexed, and querying the index with a vector of a different dimensionality returns an error. It is recommended to provide the dimensions.
+Accepted values::: `INTEGER` between `1` and `4096` inclusively.
+Default value::: None, though the setting can be omitted. The setting was mandatory prior to Neo4j 5.23.
```
JPryce-Aklundh:
```suggestion
The dimensionality of the vectors to be indexed.
For more information, see xref:indexes/semantic-indexes/vector-indexes.adoc#embeddings[].
This setting can be omitted, and any `LIST<INTEGER | FLOAT>` can be indexed and queried, separated by their dimensionality.
Setting this value adds additional checks that ensure only vectors with the configured dimensionality are indexed, and querying the index with a vector of a different dimensionality returns an error. It is recommended to provide the dimensions.
Accepted values::: `INTEGER` between `1` and `4096` inclusively.
Default value::: None.
The setting was mandatory prior to Neo4j 5.23.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,58 @@ You can also create a vector index for relationships with a particular type on a
...
+
+`vector.similarity_function` ::
+The name of the similarity function to use to assess how similar two vectors are.
+To read more about the available similarity functions, see xref:indexes/semantic-indexes/vector-indexes.adoc#similarity-functions[].
+
+Accepted values::: `STRING`: `'cosine'`, `'euclidean'`.
+Default value::: `'cosine'`. The setting was mandatory prior to Neo4j 5.23.
```
JPryce-Aklundh:
```suggestion
`vector.similarity_function` ::
The name of the similarity function used to assess the similarity of two vectors.
To read more about the available similarity functions, see xref:indexes/semantic-indexes/vector-indexes.adoc#similarity-functions[].

Accepted values::: `STRING`: `'cosine'`, `'euclidean'`.
Default value::: `'cosine'`. The setting was mandatory prior to Neo4j 5.23.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,58 @@ You can also create a vector index for relationships with a particular type on a
...
+
+`vector.quantization.enabled` label:new[Introduced in 5.23] ::
+Quantization is a technique to reduce the size of vector representations.
+Enabling quantization can accelerate search performance but can slightly decrease accuracy.
+This is recommended to always be enabled on machines with limited memory.
+Vector indexes created prior to Neo4j 5.23 have this setting effectively set as `false`.
+
+Accepted values::: `BOOLEAN`: `true`, `false`.
+Default value::: `true`
```
JPryce-Aklundh:
```suggestion
`vector.quantization.enabled` label:new[Introduced in 5.23] ::
Quantization is a technique to reduce the size of vector representations.
Enabling quantization can accelerate search performance but can slightly decrease accuracy.
It is recommended to enable quantization on machines with limited memory.
Vector indexes created prior to Neo4j 5.23 have this setting effectively set to `false`.

Accepted values::: `BOOLEAN`: `true`, `false`.
Default value::: `true`
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,58 @@ You can also create a vector index for relationships with a particular type on a
...
+
+==== Advanced configuration settings
```
JPryce-Aklundh: Maybe we could call this "HNSW configuration settings" and have a sentence underneath it explaining abbreviation and why it is used for these 2 settings?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,58 @@ You can also create a vector index for relationships with a particular type on a
...
+
+`vector.hnsw.m` label:new[Introduced in 5.23] ::
+The `M` parameter controls the maximum number of connections each node has in the HNSW graph.
+Increasing this value may lead to greater accuracy at the expense of increased index population and update times, especially for vectors with high dimensionality.
+Vector indexes created prior to Neo4j 5.23 have this setting effectively set as `16`.
```
JPryce-Aklundh:
```suggestion
Vector indexes created prior to Neo4j 5.23 have this setting effectively set to `16`.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,58 @@ You can also create a vector index for relationships with a particular type on a
...
+
+Accepted values::: `INTEGER` between `1` and `512` inclusively.
+Default value::: `16`
+
+`vector.hnsw.ef_construction` label:new[Introduced in 5.23] ::
+The number of nearest neighbors tracked during the insertion of vectors into the HNSW graph.
+Increasing this value increases the quality of the index, and may lead to greater accuracy (with diminishing returns) at the expense of increased index population and update times.
```
JPryce-Aklundh, line 188: Can you explain what "diminishing returns" means here?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -117,12 +119,58 @@ You can also create a vector index for relationships with a particular type on a
...
+Vector indexes created prior to Neo4j 5.23 have this setting effectively set as `100`.
```
JPryce-Aklundh:
```suggestion
Vector indexes created prior to Neo4j 5.23 have this setting effectively set to `100`.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -463,18 +509,32 @@ The following table lists the known issues and, if fixed, the version in which t
 |===
 | Known issues | Fixed in
 
+| Using the legacy procedure link:{neo4j-docs-base-uri}/operations-manual/{page-version}/reference/procedures/#procedure_db_index_vector_createnodeindex[`db.index.vector.createNodeIndex`] on a database created prior to Neo4j 5.11 in Neo4j 5.18 and later, before the database has been upgraded, can cause an error.
```
JPryce-Aklundh: "...on a database created prior to Neo4j 5.11 in Neo4j 5.18..." is confusing to me, and Im not sure I understand it well enough to offer any suggestion.
Does it mean something like "...on a database created with a Neo4j version earlier than 5.11, while running Neo4j 5.18 or later, can cause an error if the database has not been upgraded"?

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -463,18 +509,32 @@ The following table lists the known issues and, if fixed, the version in which t
...
+In Neo4j 5.20, the error was clarified.
+[TIP]
+--
+Using the `CREATE VECTOR INDEX` command instead avoids this issue and also triggers the upgrade.
+
+If the use of the procedure is out of the user's control, trigger an upgrade.
+The simplest is to write to the database on the newer binary.
+An alternative is to use link:{neo4j-docs-base-uri}/operations-manual/{page-version}/tools/neo4j-admin/migrate-database/[`neo4j-admin database migrate`] if you have direct access to the machine running the Neo4j DBMS.
```
JPryce-Aklundh: What upgrade are we talking about here? Neo4j DBMS? 
I find this section a bit confusing, especially "write to the database on the newer binary".
Does it mean basically, that if you have to (for whatever reason) continue using the procedure, you need to manually trigger an upgrade to the latest version of Neo4j, and once upgraded, you can perform your operations?

JPryce-Aklundh (commented):
Thanks @parnmatt! I have some editorial suggestions, and a few areas where I didn't fully understand what was going on. Also, we need to add the 3 new settings to the Additions and Deprecations page (I can do this once we have clarified everything else, if you like).

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: I would be open to that; however, it should still be made clear to the user that these are "advanced" tuning settings.
These index settings are arguably more for those who already know what they are.
I think somewhere in the docs I've seen something referred to as "advanced" user settings/config or something.

How would you suggest we handle that "theme".

parnmatt: What do you mean by 'explaining abbreviation'?
Do you mean the abbreviation of HNSW? I believe that _was_ in the docs, though it may have been moved, or we only reference the paper now.

Or do you mean why they are called `...m` and `...ef_construction`?
`M` and `efConstuction` are the names in the paper, and the names most vendors use. Their chosen names are often not expanded further and seem 'understood' in that space. Hence most describe what they mean and do, rather than rationalise the name :sweat_smile: 
The paper defines:
> number of established connections _M_, […] size of the dynamic candidate list _efConstruction_,

It's common to use `M` in maths/algorithms for a "number of xxx" when `N` is not conventionally used.
Rarely `ef` has been explicitly discussed, in one small library there's a comment suggesting they think it stands for _effervescence coefficient_ … which only loosely makes sense in regards to the paper's description. In another place, I've seen it loosely implying it stands for _exploritory factor_. That's it.
`efConstruction` is `ef` during construction, and `efSearch` is `ef` during searching.

It may be named more explicitly if I follow a few paper references …but I don't think it's important as no one else seems to document it.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: It's quite a normal phrase. Though I was struggling where to exactly put it in the sentence.

Basically, increasing `efConstruction` makes a larger change to accuracy initially; however, that perceived change diminishes as you keep increasing it; especially as it gets more expensive as you increase it. The value to time ratio decreases.

Pulling 'random' numbers:
| input | output |
| - | - |
| 1 | 10.00 |
| 2 | 20.00 |
| 3 | 25.00 |
| 4 | 27.50 |
| 5 | 28.75 |

This sequence has diminishing returns. As we increase, the relative difference decreases. If we then increase the time it takes to get a output … then it really gets to the point of "is it worth to increase anymore".

In the space of vector indexing; depending on the data, it may be beneficial to increase this a little, or a lot, to get the same wanted accuracy. Or perhaps a particular use case doesn't care about time and just wants really accurate results and is willing to pay that expense. Others may only want a brief idea, thus can be inaccurate, and may even turn it down!

A few other venders note this, and also use "diminishing returns"
I believe it is important to note something like this, ideally with that or similar phrasing; otherwise a user may just configure it to the max and then get something extremely expensive, and very little difference in accuracy compared to a much smaller value.

Again, these are "advanced" settings.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: It's not the most clear I agree.

Ok, I will expand the details here, and lets see if we can find a simple and clear way to set the scene.

(Upgrade in a nutshell, for reference)

There are 3 versions to think of.
* The binary version, the version of the JAR downloaded/compiled.
* The runtime version, the version the DBMS is running as.
* The kernel version, the version of the database

During an upgrade (think cluster), we are switching out binaries to newer ones (binary). During that time, the machines on the new binaries cannot use the new stuff, they have to talk in the older version that the others are on (runtime). Once all machines are on the new binary, we let it know it can update to the newer runtime (this was an explicit procedure call, but I think recent clusters can figure this out now, not sure). Once the system knows it's now safe to talk on the new things and use new things, the first write operation to a database will upgrade the database to the new version (kernel). So even though the database "older", because the "runtime" is newer, we can use the new things as that operation will update the database.

This also happens in community edition, but is a little simpler.

---

Ok with that context.

Someone creates a Neo4j database on a version prior to 5.11. Lets say on a binary version of 5.7 for an example.
They populate the database with their data. Perhaps they also fill it with pregenerated embeddings.

They want to use Neo4j vector stuff, so they update their binaries. They pick a new version (5.18 or newer) lets say they grab 5.22 as that's the newest.

The start the db, and they may optionally do a few read queries to check things are still working, but do not do a write. The first write operation they do is to create a vector index using the legacy procedure. _*BOOM*_ error.
In this case, the error is because in that procedure it will check what that _kernel_ version can do. 5.7 cannot do vectors.

parnmatt: How about something more like:

```suggestion
| The creation of a vector index using the legacy procedure link:{neo4j-docs-base-uri}/operations-manual/{page-version}/reference/procedures/#procedure_db_index_vector_createnodeindex[`db.index.vector.createNodeIndex`] may fail with an error in Neo4j 5.18 and later if the database was last written to with a version prior to Neo4j 5.11, and the legacy procedure is the first write operations.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: I'm really not a fan of this part.
Basically yes, we need to trigger the kernel version upgrade to the newer runtime version, in line with the binary version. There are a few ways to do it.

There is something in the procedure that will fail when it tries to figure out what vector index versions it has access to. It reads the kernel version and finds a version before we support vector index … and it doesn't like that.
Really it needs to be the runtime version here; however, we didn't expose that in the procedure framework and its debatable if we want to.

parnmatt: Though very effective, we should avoid `neo4j-admin database migrate` … it's not advised to do this manually in a clustered environment outside of an actual migration and re-seed.

So the options are:
* Use the cypher command that replaced the legacy procedure in 5.15
* Do _any-other_ write operation to the database first.

I mention the latter, as it is a surefire way to ensure the write transaction. Especially if someone is using a framework that is stuck using the procedure. I think LangChain was using it for quite a while … I hope they've moved on.

parnmatt: Thoughts on:

```suggestion
Using the `CREATE VECTOR INDEX` command instead avoids this issue.
If the use of the procedure is out of the user's control, any other write operation to the database on the newer binary before using the procedure will avoid the issue.
```

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: Ok, I see. Let's go with "Advanced settings", then. We removed mentions of "HNSW" when we did the rewrite a few months ago as we thought it was too academic a term. But, as you say, these are advanced settings, so Im fine with its mention here. Perhaps the first time the abbreviation is mentioned, we could just bracket its meaning (e.g. "...each node has in the HNSW (Hierarchical Navigable Small Worlds) graph")

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: Ok. I know the meaning of the phrase, but was unsure about its precise meaning in this context, so thank you for explaining. I think it is probably fine as it stands 👍

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: I like your suggestion, but could we modify it as follows:
"If the use of the procedure is unavoidable, performing any other write operation to the database on the newer binary before using the procedure will avoid the issue."

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
JPryce-Aklundh: I like it. Maybe the last bit could say
"...and later if the database was last written to with a version prior to Neo4j 5.11, and the legacy procedure is the first write operation used on the newer version."

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: Sorry, I wasn't sure 😅

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: 👍 so:

```suggestion
| The creation of a vector index using the legacy procedure link:{neo4j-docs-base-uri}/operations-manual/{page-version}/reference/procedures/#procedure_db_index_vector_createnodeindex[`db.index.vector.createNodeIndex`] may fail with an error in Neo4j 5.18 and later if the database was last written to with a version prior to Neo4j 5.11, and the legacy procedure is the first write operation used on the newer version.
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: 👍 so:
```suggestion
Using the `CREATE VECTOR INDEX` command instead avoids this issue.
If the use of the procedure is unavoidable, performing any other write operation to the database on the newer binary before using the procedure will avoid the issue
```

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
parnmatt: Sounds good, I'll have a look

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Deprecations, additions, and compatibility](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/deprecations-additions-removals-compatibility)
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)
- [Syntax](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/syntax)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Deprecations, additions, and compatibility](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/deprecations-additions-removals-compatibility)
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)
- [Syntax](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/syntax)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Deprecations, additions, and compatibility](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/deprecations-additions-removals-compatibility)
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)
- [Syntax](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/syntax)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Deprecations, additions, and compatibility](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/deprecations-additions-removals-compatibility)
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)
- [Syntax](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/syntax)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1009.surge.sh

Updated pages:
- [Deprecations, additions, and compatibility](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/deprecations-additions-removals-compatibility)
- [Vector indexes](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/semantic-indexes/vector-indexes)
- [Syntax](https://neo4j-docs-cypher-1009.surge.sh/cypher-manual/5/indexes/syntax)

neo-technology-commit-status-publisher:

PR #1147, opened 2025-01-08
The Cypher query option `eagerAnalyzer` is deprecated

blouerat:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1147.surge.sh

Updated pages:
- [Deprecations, additions, and compatibility](https://neo4j-docs-cypher-1147.surge.sh/cypher-manual/2025/deprecations-additions-removals-compatibility)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1147.surge.sh

Updated pages:
- [Deprecations, additions, and compatibility](https://neo4j-docs-cypher-1147.surge.sh/cypher-manual/2025/deprecations-additions-removals-compatibility)

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -16,6 +16,36 @@ New features are added to the language continuously, and occasionally, some feat
...
+----
+
+[source, cypher, role="noheader"]
+----
+CYPHER eagerAnalyzer=lp MATCH (a)-->(b) DELETE b RETURN a
+----
+a|
+
+The Cypher query option `eagerAnalyzer` is deprecated and will be removed without a replacement.
+The option is ignored, eagerness analysis is systematically performed on the logical plan regardless of the value provided.
```
```suggestion
This is because xref:planning-and-tuning/execution-plans.adoc#lazy-eager-evaluation[eagerness analysis] is systematically performed on the logical plan regardless of any value provided.
```

JPryce-Aklundh (approved):
Looks good, just 2 editorial suggestions

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
blouerat: This isn't quite right I'm afraid.
As described in the section you linked, Cypher is executed lazily as much as possible, processing one row (or batch of rows) at a time instead of materialising each intermediate result. Some operations like counting or sorting (in certain scenarios) do need to process all the intermediate results, that's why you end up with plans like `EagerAggregation`.
_Eagerness analysis_, probably a confusing name, is a different beast though it is related. Another place where laziness might be a problem is when the query contains both reads and writes. We need to make sure that a read clause can't see the effect of a subsequent write clause, and that it can always see the effect of a precedent write clause. This sometimes requires materialising the intermediate results using [the Eager operator](https://neo4j.com/docs/cypher-manual/current/planning-and-tuning/operators/operators-detail/#query-plan-eager). This is what eagerness analysis is about: inserting Eager where relevant in the context of a read/write query.

---

The initial implementation was based on the intermediate representation of the query (IR) but we replaced it with a new one which is based on logical plans (LP). LP has been the default implementation since the summer, and we're now confident enough to get rid of the IR implementation.

blouerat: Going for:

> The Cypher query option `eagerAnalyzer` is deprecated and will be removed without a replacement.
> Eagerness analysis is systematically performed on the logical plan regardless of the value provided.

as discussed separately.

neo-technology-commit-status-publisher:

PR #1151, opened 2025-01-09
Fix examples in Patterns / Syntax and semantics / shortestPaths

dogofbrian:
The examples in Patterns / Syntax and semantics / shortestPaths / Examples mix legacy shortest path and QPPs, which is disallowed. This PR changes them to variable-length relationships.

It also improves the header "Restricted to variable length". QPPs are disallowed but also variable length. What they are not are variable-length relationships.


neo-technology-commit-status-publisher:

JPryce-Aklundh (approved):

PR #1152, opened 2025-01-09
Explain trigram indexing for text-2.0 provider

JPryce-Aklundh:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1152.surge.sh

Updated pages:
- [Create, show, and delete indexes](https://neo4j-docs-cypher-1152.surge.sh/cypher-manual/2025/indexes/search-performance-indexes/managing-indexes)
- [The impact of indexes on query performance](https://neo4j-docs-cypher-1152.surge.sh/cypher-manual/2025/indexes/search-performance-indexes/using-indexes)

modules/ROOT/pages/indexes/search-performance-indexes/managing-indexes.adoc
```diff
@@ -286,6 +286,18 @@ See the section about xref:indexes/search-performance-indexes/using-indexes.adoc
...
 Text indexes are only used for exact query matches. To perform approximate matches (including, for example, variations and typos), and to compute a similarity score between `STRING` values, use semantic xref:indexes/semantic-indexes/full-text-indexes.adoc[full-text indexes] instead.
 
+[[text-indexes-trigram-indexes]]
+==== Trigram indexing
+
+The default text index provider, `text-2.0`, uses trigram indexing.
+This means that `STRING` values are indexed into overlapping trigrams, each containing three characters.
+For example, the word `"developer"` would be indexed by the following trigrams: `["dev", "eve", "vel", "elo", "lop", "ope", "per"]`.
+
+This makes text indexes particularly suited for substring (`CONTAINS`) and suffix (`ENDS WITH`) searches.
```
parnmatt: may want to make a note something _as well as_ `STARTS WITH` (which is all that `Range` can handle)

modules/ROOT/pages/indexes/search-performance-indexes/managing-indexes.adoc
```diff
@@ -286,6 +286,18 @@ See the section about xref:indexes/search-performance-indexes/using-indexes.adoc
...
+For example, searches like `CONTAINS "vel"` or `ENDS WITH "per"` can be efficiently performed by directly looking up the relevant trigrams in the index.
+By comparison, range indexes, which indexes `STRING` values alphabetically (see xref:indexes/search-performance-indexes/using-indexes.adoc#range-index-backed-order-by[Range index-backed `ORDER BY`] for more information) and are therefore more suited for prefix searches (`STARTS WITH`), would need to scan through all indexed values to check if `"vel"` existed anywhere within the text.
```
parnmatt: Depends how correct you want to be here. It's not ordered alphabetically, as strings are more than just the alphabet, and the alphabet order is different depending on the language.

What is often used is some lexographical ordering based on some representation (like unicode code point). This is more general and covers all strings.

However, specifically for us in the Range index, we order the strings lexographically, byte-wise, in its UTF-8 representation.
How specific you want to be there, is up to you.

modules/ROOT/pages/indexes/search-performance-indexes/managing-indexes.adoc
```diff
@@ -286,6 +286,18 @@ See the section about xref:indexes/search-performance-indexes/using-indexes.adoc
 [TIP]
 Text indexes are only used for exact query matches. To perform approximate matches (including, for example, variations and typos), and to compute a similarity score between `STRING` values, use semantic xref:indexes/semantic-indexes/full-text-indexes.adoc[full-text indexes] instead.
 
+[[text-indexes-trigram-indexes]]
+==== Trigram indexing
+
+The default text index provider, `text-2.0`, uses trigram indexing.
+This means that `STRING` values are indexed into overlapping trigrams, each containing three characters.
```
parnmatt: To be specific here, it is not "three characters" (though that is the case for ascii values), it's three unicode codepoints.
_("characters" is also an odd one, as it differs in meaning depending on context)_

An example to show the subtle difference, `"Malmö"` would be `["Mal", "alm", "lmö"]` because `ö` is a canonical composite, thus a single code point. However, `"Malmö"` is a _different_ string, the `ö` is actually `o + ◌̈`, it is two code points; thus  `"Malmö"` would be `["Mal", "alm", "lmo", mö]`.
_It gets worse with "characters" composed of _multiple_ combination marks._

The strings `"Malmö"` and `"Malmö"` are canonically equivalent in unicode, and rendered the same. But those strings do have different code points.

Some systems do a normalisation of the strings, there are a few different kinds. I do not believe we do so and just take the data the user gives us as is.

I doubt it would be needed in such detail; I believe the current explanation is sufficient. However, it would be best to refer to **Unicode codepoints** rather than _characters_.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-cypher-1152.surge.sh

Updated pages:
- [Create, show, and delete indexes](https://neo4j-docs-cypher-1152.surge.sh/cypher-manual/2025/indexes/search-performance-indexes/managing-indexes)
- [The impact of indexes on query performance](https://neo4j-docs-cypher-1152.surge.sh/cypher-manual/2025/indexes/search-performance-indexes/using-indexes)

parnmatt (approved):

neo-technology-commit-status-publisher:

PR #1155, opened 2025-01-13
Change page-version attribute in external links to current

JPryce-Aklundh:
With the release of Cypher 25, the `page-version` attribute in Cypher and server doc sets will differ. It will be replaced with `current`.


neo-technology-commit-status-publisher:

PR #1316, opened 2025-06-23
Mark 25 as prerelease - force sitemap to Cypher 5

recrwplay:
Marking 25 as prerelease will do the following:

- Ensure the sitemaps extension treats Cypher 5 as the latest version, so that it will generate a sitemap from Cypher 5 content
- Ensure canonical URLs in the HTML `<head>` are accurate. We can then allow new pages in Cypher 25 to be indexed, or we can index the whole of the Cypher manual.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -1,6 +1,7 @@
 name: cypher-manual
 title: Cypher Manual
 version: '25'
+prerelease: true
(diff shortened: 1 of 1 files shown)


neo4j-docops-agent:

JPryce-Aklundh (approved):

PR #1343, opened 2025-07-17
Name Repeat operator without Trail/Walk parameter

arnefischereit:
All the description of the operator should be applicable independent of the traversal path mode. In EXPLAIN outputs, the parameter will be included, though.

I have left the description in the additions document, as this is how it was introduced. Maybe we should rename it there as well.


Dunfluff (approved):
LGTM

rsill-neo4j (approved):
to which version(s) does this apply?

arnefischereit:
I would argue it should have been like this since the repeat operator was introduced. But it becomes most relevant from Cypher version 25, when we have `REPEATABLE ELEMENTS`.

neo4j-docops-agent:

PR #1435, opened 2025-12-04
Removed possibly broken syntax

JoelBergstrand:
The returning of a constant in a NEXT query is non-idomatic and might be broken syntax that should be removed.


neo4j-docops-agent:

rsill-neo4j (approved):

PR #1445, opened 2025-12-22
Fix query plans for `Anti*` operators

stefano-ottolenghi:
When finalized, changes apply to cypher 5 too.


neo4j-docops-agent:

PR #1464, opened 2026-02-02
SURF-243 Document precedence rules in Cypher

neo-harshg:
https://linear.app/neo4j/issue/SURF-243/document-precedence-rules-in-cypher


neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1464.surge.sh

Updated pages:
- [Boolean operators](https://neo4j-docs-cypher-1464.surge.sh/cypher-manual/25/expressions/predicates/boolean-operators)

modules/ROOT/pages/expressions/predicates/boolean-operators.adoc
```diff
@@ -137,6 +137,30 @@ RETURN n.name AS name, n.age AS age
...
+* `AND`
+* `XOR`
+* `OR`
+
+This means `NOT` applies to the result of a comparison unless you add parentheses.
+
+Examples:
+
+* `NOT true = false` is parsed as `NOT (true = false)` and is valid.
+* `true = NOT false` does not parse. Use `true = (NOT false)` if that is your intent.
```
rsill-neo4j: when it doesn't parse, does that mean it's not valid?

rsill-neo4j:
which release(s) does this apply to?

modules/ROOT/pages/expressions/predicates/boolean-operators.adoc
```diff
@@ -137,6 +137,30 @@ RETURN n.name AS name, n.age AS age
 2+d|Rows: 4
 |===
 
+.Operator precedence
+[NOTE]
+====
+Operator precedence defines how Cypher groups expressions when parentheses are not present.
+Within predicates, the following order is relevant (higher precedence first):
+
+* Comparison operators: `=`, `<>`, `<`, `>`, `<=`, `>=`, `IS NULL`, `IS NOT NULL`, `IN`, `STARTS WITH`, `ENDS WITH`, `CONTAINS`, `=~`
```
hvub, line 146: `<=` gets rendered as a left arrow. 

Image

Hopefully, that can be prevented somehow. Try:

```suggestion
* Comparison operators: `=`, `<>`, `<`, `>`, `\<=`, `>=`, `IS NULL`, `IS NOT NULL`, `IN`, `STARTS WITH`, `ENDS WITH`, `CONTAINS`, `=~`
```

modules/ROOT/pages/expressions/predicates/boolean-operators.adoc
hvub: Yes. I agree with what is presumably behind your question, which is that users do not care where a query fails in the stack. From a user's point of view it only matter whether a query is valid or invalid.

I would also move the invalid example to the back so it is not a back and forth between valid and invalid examples.

```
* `NOT true AND false` is valid and equivalent to `(NOT true) AND false`.
* `true AND NOT false` is valid and equivalent to `true AND (NOT false)`.
* `NOT true = false` is valid and equivalent to `NOT (true = false)`.
* `true = NOT false` is not valid. Use `true = (NOT false)` instead.
```

hvub (commented):
Apart from the detailed comment, I think this okay and certainly a helpful addition to the docs.

I had hoped though that the docs would gain a full list of precedence of the Cypher expressions and the back of this issue. I am thinking of a dedicated page titled "Precedence" or "Expression precedence" as a subitem under "Expression", last after "Conditional expressions (CASE)".

@rsill-neo4j you should weigh in here, if you would like such a page or not.

The NOTE as done be this PR could stay, maybe shorten a bit in the first paragraph in favor of linking to the dedicated page.

I have linked to documents on the issues that given an overview of Cypher precedence, at least as of 2023. To google doc is probably the best source. The terminology use there probably need alignment to fit with the docs. Also linking and use of concrete syntax would make a better docs fit.

hvub:
> which release(s) does this apply to?

I think all. It just documents something that was not explicitly documented before, i.e. it also applies to Cypher 5 and Cypher 25 alike.

If we go ahead it a full precedence documentation the answer will be the same, except that we will have to make minor adjustment for new feature... and I think there is one precedence change from Cypher 5 to Cypher 25 w.r.t. the label expression predicate. I will link the CIP on the issue.

rsill-neo4j:
> Apart from the detailed comment, I think this okay and certainly a helpful addition to the docs.
> 
> I had hoped though that the docs would gain a full list of precedence of the Cypher expressions and the back of this issue. I am thinking of a dedicated page titled "Precedence" or "Expression precedence" as a subitem under "Expression", last after "Conditional expressions (CASE)".
> 
> @rsill-neo4j you should weigh in here, if you would like such a page or not.
> 
> The NOTE as done be this PR could stay, maybe shorten a bit in the first paragraph in favor of linking to the dedicated page.
> 
> I have linked to documents on the issues that given an overview of Cypher precedence, at least as of 2023. To google doc is probably the best source. The terminology use there probably need alignment to fit with the docs. Also linking and use of concrete syntax would make a better docs fit.

sounds reasonable to add such a page

modules/ROOT/pages/expressions/predicates/boolean-operators.adoc
```diff
@@ -137,6 +137,30 @@ RETURN n.name AS name, n.age AS age
...
+* `OR`
+
+This means `NOT` applies to the result of a comparison unless you add parentheses.
+
+Examples:
+
+* `NOT true = false` is parsed as `NOT (true = false)` and is valid.
+* `true = NOT false` does not parse. Use `true = (NOT false)` if that is your intent.
+* `true AND NOT false` is parsed as `true AND (NOT false)`.
+* `NOT true AND false` is parsed as `(NOT true) AND false`.
```
rsill-neo4j:
```suggestion
* `NOT true AND false` is valid and equivalent to `(NOT true) AND false`.
* `true AND NOT false` is valid and equivalent to `true AND (NOT false)`.
* `NOT true = false` is valid and equivalent to `NOT (true = false)`.
* `true = NOT false` is not valid. Use `true = (NOT false)` instead.
```

suggestion following @hvub 's proposal

neo4j-docops-agent:

rsill-neo4j (approved):

PR #1479, opened 2026-02-10
SURF-446 Add doc for new predicate PROPERTY_EXISTS

neo-harshg:
As part of making cypher GQL compliant we have introduced PROPERTY_EXISTS predicate.

Details: https://linear.app/neo4j/issue/SURF-446/gql-compliance-property-exists-g115

### DIFF
--- a/modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
+++ b/modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
@@ -89,6 +89,13 @@ These codes order the features in the table below.
 | xref:patterns/reference.adoc#label-expressions[Label expressions]
 |
 
+| [[g115]]G115
+| Property exists predicate
+| xref:expressions/predicates/comparison-operators.adoc#property-exists-predicate[`PROPERTY_EXISTS`]
+| GQL allows testing whether a value exists and is not null.
+In Cypher, the same semantics are expressed with the xref:expressions/predicates/comparison-operators.adoc[`IS NOT NULL`] predicate: for nodes and relationships, a missing property evaluates to `null`, so `n.prop IS NOT NULL` is true when the property is present and not null.
+As of Neo4j 2026.03, the same semantics can also be expressed with the `PROPERTY_EXISTS(variable, propertyName)` predicate (e.g. `WHERE PROPERTY_EXISTS(n, name)`).
+
 | GA06
 | Value type predicates
 | xref:expressions/predicates/type-predicate-expressions.adoc[Type predicate expressions]
--- a/modules/ROOT/pages/expressions/predicates/comparison-operators.adoc
+++ b/modules/ROOT/pages/expressions/predicates/comparison-operators.adoc
@@ -197,6 +197,7 @@ RETURN n.name AS name
 1+d|Rows: 2
 |===
 
+[[is-not-null-operator]]
 .`IS NOT NULL` operator
 // tag::expressions_predicates_comparison_operators_is_not_null[]
 [source, cypher]
@@ -220,7 +221,37 @@ RETURN n.name AS name, n.email AS email
 2+d|Rows: 4
 |===
 
+The `IS NOT NULL` predicate tests whether an expression evaluates to a non-`NULL` value.
+For the GQL property existence predicate (GQL feature G115), see xref:property-exists-predicate[].
+
 =====
+[[property-exists-predicate]]
+.`PROPERTY_EXISTS` predicate
+The optional GQL feature G115 defines a `PROPERTY_EXISTS(element, propertyName)` predicate for testing whether a property exists on a node or relationship and is not `NULL`.
+
+As of Neo4j 2026.03, Cypher supports the GQL syntax:
+
+[source, cypher]
+----
+MATCH (n:Person)
+WHERE PROPERTY_EXISTS(n, email)
+RETURN n.name AS name, n.email AS email
+----
+
+For node and relationship properties, this is equivalent to:
+
+[source, cypher]
+----
+MATCH (n:Person)
+WHERE n.email IS NOT NULL
+RETURN n.name AS name, n.email AS email
+----
+
+Notes and differences:
+
+* The second argument must be a property name identifier (for example, `email`). String literals and expressions are not allowed.
+* If the element argument is `NULL`, `PROPERTY_EXISTS` returns `NULL` (consistent with GQL semantics).
+* `PROPERTY_EXISTS` applies to node and relationship properties only. `IS NOT NULL` is more general and can be applied to any expression.
 
 [[chaining-comparison-oeprators]]
 == Chaining comparison operators
--- a/modules/ROOT/pages/expressions/predicates/comparison-operators.adoc
+++ b/modules/ROOT/pages/expressions/predicates/comparison-operators.adoc
@@ -12,7 +12,7 @@ Cypher contains the following comparison operators:
 * Less than or equal to: `\<=`
 * Greater than or equal to: `>=`
 * `IS NULL`
-* `IS NOT NULL`
+* `IS NOT NULL` as well as the xref:property-exists-predicate[]
 
 [NOTE]
 For more information about how Cypher orders and compares different value types, see xref:values-and-types/ordering-equality-comparison.adoc[Values and types -> Equality, ordering, and comparison of value types]
@@ -225,8 +225,10 @@ The `IS NOT NULL` predicate tests whether an expression evaluates to a non-`NULL
 For the GQL property existence predicate (GQL feature G115), see xref:property-exists-predicate[].
 
 =====
+
 [[property-exists-predicate]]
-.`PROPERTY_EXISTS` predicate
+[discrete]
+=== `PROPERTY_EXISTS` predicate
 The optional GQL feature G115 defines a `PROPERTY_EXISTS(element, propertyName)` predicate for testing whether a property exists on a node or relationship and is not `NULL`.
 
 As of Neo4j 2026.03, Cypher supports the GQL syntax:
--- a/modules/ROOT/pages/values-and-types/working-with-null.adoc
+++ b/modules/ROOT/pages/values-and-types/working-with-null.adoc
@@ -98,5 +98,6 @@ a[coalesce($lower,0)..coalesce($upper,size(a))]
 [[is-null-is-not-null]]
 == Using `IS NULL` and `IS NOT NULL`
 Testing any value against `null`, either with the `=` operator or with the `<>` operator, always evaluates to `null`.
-Therefore,  use the special equality operators xref:expressions/predicates/comparison-operators.adoc[`IS NULL` or `IS NOT NULL`].
+Therefore, use the special equality operators xref:expressions/predicates/comparison-operators.adoc[`IS NULL` or `IS NOT NULL`].
+Using `IS NOT NULL` on a property (e.g. `n.prop IS NOT NULL`) tests that the property exists and is not null.
 
--- a/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
+++ b/modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
@@ -23,6 +23,26 @@ Cypher 25 was introduced in Neo4j 2025.06 and can only be used on Neo4j 2025.06+
 Features removed in Cypher 25 are still available on Neo4j 2025.06+ databases either by prepending a query with `CYPHER 5` or by having Cypher 5 as the default language for the database.
 For more information, see xref:queries/select-version.adoc[].
 
+[[cypher-deprecations-additions-removals-2026.03]]
+== Neo4j 2026.03
+
+=== New in Cypher 25
+
+[cols="2", options="header"]
+|===
+| Feature
+| Details
+
+a|
+label:functionality[]
+label:new[]
+
+`PROPERTY_EXISTS` predicate
+
+| A new `PROPERTY_EXISTS(element, propertyName)` predicate had been added for testing whether a property exists on a node or relationship and is not `NULL`.
+
+|===
+
 [[cypher-deprecations-additions-removals-2026.02]]
 == Neo4j 2026.02
 


neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1479.surge.sh

Updated pages:
- [Supported optional GQL features](https://neo4j-docs-cypher-1479.surge.sh/cypher-manual/25/appendix/gql-conformance/supported-optional)
- [Comparison operators](https://neo4j-docs-cypher-1479.surge.sh/cypher-manual/25/expressions/predicates/comparison-operators)
- [Working with <code>null</code>](https://neo4j-docs-cypher-1479.surge.sh/cypher-manual/25/values-and-types/working-with-null)

neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1479.surge.sh

Updated pages:
- [Supported optional GQL features](https://neo4j-docs-cypher-1479.surge.sh/cypher-manual/25/appendix/gql-conformance/supported-optional)
- [Comparison operators](https://neo4j-docs-cypher-1479.surge.sh/cypher-manual/25/expressions/predicates/comparison-operators)
- [Working with <code>null</code>](https://neo4j-docs-cypher-1479.surge.sh/cypher-manual/25/values-and-types/working-with-null)

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
```diff
@@ -89,6 +89,14 @@ These codes order the features in the table below.
 | xref:patterns/reference.adoc#label-expressions[Label expressions]
 |
 
+| [[g115]]G115
+| Property existence predicate
```
rsill-neo4j: is this the PROPERTY_EXISTS predicate? because it's only mentioned in the PR title

rsill-neo4j (commented):
the predicate is new in GQL?


is this something we should add to the additions page?

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
```diff
@@ -89,6 +89,15 @@ These codes order the features in the table below.
 | xref:patterns/reference.adoc#label-expressions[Label expressions]
 |
 
+| [[g115]]G115
+| Property exists predicate
+| xref:expressions/predicates/comparison-operators.adoc#is-not-null-operator[`IS NOT NULL`], xref:values-and-types/working-with-null.adoc#is-null-is-not-null[Working with null]
+| GQL allows testing whether a value (typically a property) exists and is not null.
+In Cypher, the same semantics are expressed with the xref:expressions/predicates/comparison-operators.adoc[`IS NOT NULL`] predicate: for nodes and relationships, a missing property evaluates to `null`, so `n.prop IS NOT NULL` is true when the property is present and not null.
+In Cypher 25, the same semantics can also be expressed with the `PROPERTY_EXISTS(variable, propertyName)` predicate (e.g. `WHERE PROPERTY_EXISTS(n, name)`), which is a standalone predicate and cannot be combined with other expressions in a single expression.
```
rsill-neo4j: which Cypher release introduced PROPERTY_EXISTS ?

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
```diff
@@ -89,6 +89,15 @@ These codes order the features in the table below.
 | xref:patterns/reference.adoc#label-expressions[Label expressions]
 |
 
+| [[g115]]G115
+| Property exists predicate
+| xref:expressions/predicates/comparison-operators.adoc#is-not-null-operator[`IS NOT NULL`], xref:values-and-types/working-with-null.adoc#is-null-is-not-null[Working with null]
+| GQL allows testing whether a value (typically a property) exists and is not null.
```
Lojjs: Regarding the "typically a property" part, this make it sound as it can be used for other values than properties also.

But I though the GQL PROPERTY_EXISTS was for properties specifically, as opposed to IS NOT NULL which kan also be used for e.g. values in a map?

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
```diff
@@ -89,6 +89,15 @@ These codes order the features in the table below.
...
+In Cypher, the same semantics are expressed with the xref:expressions/predicates/comparison-operators.adoc[`IS NOT NULL`] predicate: for nodes and relationships, a missing property evaluates to `null`, so `n.prop IS NOT NULL` is true when the property is present and not null.
+In Cypher 25, the same semantics can also be expressed with the `PROPERTY_EXISTS(variable, propertyName)` predicate (e.g. `WHERE PROPERTY_EXISTS(n, name)`), which is a standalone predicate and cannot be combined with other expressions in a single expression.
+This can be used in xref:clauses/where.adoc[`WHERE`] clauses, xref:clauses/filter.adoc[`FILTER`] clauses, and xref:schema/constraints/create-constraints.adoc#create-property-existence-constraints[property existence constraints] (`REQUIRE ... IS NOT NULL`).
```
Lojjs: This sentence seems wrong. 

I guess it is referring to PROPERTY_EXISTS and not IS NOT NULL, as that is what the sentence just above is about? In that case it is wrong that it can be used in the property existence constraint syntax.

Either way, both IS NOT NULL and PROPERTY_EXISTS can also be used in other clauses. I don't know the full list of clauses on top of my head but at least WITH and RETURN also, e.g.

```
CYPHER 25
 MATCH (n)
 RETURN PROPERTY_EXISTS(n, prop) AS e
 ```

File: modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc (hunk 2 above)
Comment (Lojjs) [inline modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc]: What does "cannot be combined with other expressions in a single expression" mean?

This is an example of a valid query where I would maybe say that you are combining PROPERTY_EXISTS with another expression:


```
CYPHER 25
 MATCH (n)
 WHERE PROPERTY_EXISTS(n, prop) AND n.prop STARTS WITH "f"
 RETURN n
 ```

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
```diff
@@ -89,6 +89,15 @@ These codes order the features in the table below.
...
+For index use with existence checks, see xref:indexes/search-performance-indexes/using-indexes.adoc#property-existence-check[Using indexes -> Property existence checks].
```
Lojjs: I got curious and checked if we get the index benefits described in that link for PROPERTY_EXISTS in the same way as we do for IS NOT NULL. Because of the way we are rewriting it to a IS NOT NULL inside a CASE, we don't.

I don't think that is a problem implementation-wise, but I would say it means we should remove this sentence from here as it is not applicable to PROPERTY_EXISTS.

modules/ROOT/pages/expressions/predicates/comparison-operators.adoc
```diff
@@ -220,6 +221,10 @@ RETURN n.name AS name, n.email AS email
 2+d|Rows: 4
 |===
 
+The `IS NOT NULL` predicate provides the same semantics as the optional GQL property existence predicate (GQL feature G115).
+For details, see xref:appendix/gql-conformance/supported-optional.adoc#g115[GQL conformance -> Supported optional GQL features (G115)].
+In Cypher 25, the same semantics can also be expressed with the `PROPERTY_EXISTS(variable, propertyName)` predicate (e.g. `WHERE PROPERTY_EXISTS(n, name)`), which is valid only as a standalone expression.
```
Lojjs: Some feedback on this part:

- [x] It sounds like circular reasoning. The first sentence says that `IS NOT NULL` has the same semantics as G115 (which is the formal name of `PROPERTY_EXISTS`), while the second sentence say the `PROPERTY_EXISTS` has the same semantics as `IS NOT NULL`. Do you agree or am I reading it wrong?
- [x] As this branch is specifically for the Cypher 25 version of the Cypher manual, I think you do not need to mention Cypher 25 explicitly. Maybe mention which Neo4j version it was introduced in instead, like "As of Neo4j 2026.03..." 
- [x] There are some differences in semantics which were described in the CIP, mainly the difference when the node variable is null and that the expressions supports different types. I think we ought to document this. Not sure if it fits best here under `IS NOT NULL` or if we want to break out `PROPERTY_EXISTS` to its own section or if it should be on the conformance page. Maybe @rsill-neo4j has an opinion on that
- [x] I'm not sure if 'standalone expression' is a concept we use in other parts of the docs or if we need to clarify what it means for the reader.

modules/ROOT/pages/values-and-types/working-with-null.adoc
```diff
@@ -98,5 +98,6 @@ a[coalesce($lower,0)..coalesce($upper,size(a))]
 [[is-null-is-not-null]]
 == Using `IS NULL` and `IS NOT NULL`
 Testing any value against `null`, either with the `=` operator or with the `<>` operator, always evaluates to `null`.
-Therefore,  use the special equality operators xref:expressions/predicates/comparison-operators.adoc[`IS NULL` or `IS NOT NULL`].
+Therefore, use the special equality operators xref:expressions/predicates/comparison-operators.adoc[`IS NULL` or `IS NOT NULL`].
+Using `IS NOT NULL` on a property (e.g. `n.prop IS NOT NULL`) tests that the property exists and is not null, and is conformant with the optional GQL property existence predicate (xref:appendix/gql-conformance/supported-optional.adoc#g115[GQL feature G115]).
```
Lojjs: I think the first part of this sentence 

> Using `IS NOT NULL` on a property (e.g. `n.prop IS NOT NULL`) tests that the property exists and is not null

 is a good addition but I wonder if the "and is conformant..." part is a bit overkill in this place. 

The vast majority of users will not care about GQL conformance so I think we usually try to avoid referring to it from all over the docs and keep it more centered to the gql conformance page.

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
Lojjs: It should be 2026.03. I think we should indicate this instead of mention Cypher 25 explicitly (as this will anyway be included only in the Cypher 25 version of the manual). Made a comment on this in another place as well.

Lojjs (changes requested):
Some mixed feedback from me

modules/ROOT/pages/expressions/predicates/comparison-operators.adoc
rsill-neo4j: +1 for creating a section for `PROPERTY_EXISTS`

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
rsill-neo4j: ok, then i guess we can mention it also in the additions page

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
```diff
@@ -89,6 +89,13 @@ These codes order the features in the table below.
 | xref:patterns/reference.adoc#label-expressions[Label expressions]
 |
 
+| [[g115]]G115
+| Property exists predicate
+| xref:expressions/predicates/comparison-operators.adoc#is-not-null-operator[`IS NOT NULL`], xref:values-and-types/working-with-null.adoc#is-null-is-not-null[Working with null]
+| GQL allows testing whether a value (typically a property) exists and is not null.
+In Cypher, the same semantics are expressed with the xref:expressions/predicates/comparison-operators.adoc[`IS NOT NULL`] predicate: for nodes and relationships, a missing property evaluates to `null`, so `n.prop IS NOT NULL` is true when the property is present and not null.
+In Cypher 2026.03, the same semantics can also be expressed with the `PROPERTY_EXISTS(variable, propertyName)` predicate (e.g. `WHERE PROPERTY_EXISTS(n, name)`).
```
Lojjs: I think "Since Neo4j 2026.03" or "As of Neo4j 2026.03" is better than "In Cypher 2026.03" to 
1. clarify that 2026.03 is a Neo4j version and not a Cypher version
2. that is applies to 2026.03 and all following versions

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
Lojjs: @neo-harshg I think the page Richard is referring to is the one called "Additions, deprecations, removals, and compatibility". You would need to do a new table section for 2026.03 there. 

Easiest is probably to copy a "New in Cypher 25" table from further down the page and edit it.

Lojjs (commented):
Looks much better now. I have one new minor comment and two of the old ones which are not fully resolved

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
neo-harshg: applied the change.

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
neo-harshg: i have updated that page. please let me know if that is what you intended.

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
neo-harshg: Thanks for flagging it. I think because there were multiple iteration of CIP the document also had the same iteration, so it might have been a leftover content from earlier. Modified it now.

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
```diff
@@ -89,6 +89,13 @@ These codes order the features in the table below.
 | xref:patterns/reference.adoc#label-expressions[Label expressions]
 |
 
+| [[g115]]G115
+| Property exists predicate
+| xref:expressions/predicates/comparison-operators.adoc#is-not-null-operator[`IS NOT NULL`], xref:values-and-types/working-with-null.adoc#is-null-is-not-null[Working with null]
```
Lojjs: I missed this in the last review, now when you broke out PROPERTY_EXISTS to its own section, the first link here should point at that rather than IS NOT NULL section. You could also add a link to that section from the additions page (which otherwise looks great).

Lojjs (approved):
Looks good to me now

rsill-neo4j (approved):

neo4j-docops-agent:

PR #1484, opened 2026-02-17
Removed duplicate example in aggregation docs

JoelBergstrand:
This examples looks like a less developed version of the example below. I propose removing it for the example below.


neo4j-docops-agent:

JoelBergstrand:
Okey, they are not identical even though they are very similar and are titled the same.

PR #1505, opened 2026-03-24
Document path modes

rsill-neo4j:


modules/ROOT/pages/patterns/index.adoc
```diff
@@ -27,6 +27,7 @@ This chapter includes the following sections:
 * xref:patterns/shortest-paths.adoc[] - information about finding the `SHORTEST` path patterns.
 * xref:patterns/non-linear-patterns.adoc[] - information about equijoins and graph patterns (combined path patterns).
 * xref:patterns/match-modes.adoc[] - information about Cypher's different match modes, which determine whether relationships can appear more than once in a graph pattern match.
+* xref:patterns/path-modes.adoc[] - information about Cypher's different path modes, which determine whether nodes and relationships can appear more than once in a graph pattern match.
```
arnefischereit: This does not help the user decide between this point and the one above.

arnefischereit: We talk about path and graph patterns elsewhere in the documentation. Maybe we can reuse that terminology here.

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,209 @@
+:description: Information about Cypher's different path modes, which determine whether or not nodes and relationships can appear more than once in a graph pattern match.
+:page-role: new-neo4j-2026.03 cypher-25-only
+
+= Path modes
+
+Path modes determine whether or not nodes or relationships can appear more than once in a path pattern match.
```
arnefischereit: Similar to my comment in the index: Do we need to make the difference to a graph pattern match clear? Should we reference path patterns?

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,209 @@
...
+Cypher supports three path modes:
+
+* xref:patterns/match-modes.adoc#different-relationships[`WALK`]: no constraints imposed; both nodes and relationships can be repeated within the path.
+* xref:patterns/match-modes.adoc#repeatable-elements[`TRAIL`]: relationships cannot be repeated, but nodes can be repeated within the path.
+* xref:patterns/match-modes.adoc#repeatable-elements[`ACYCLIC`]: nodes cannot be repeated within a path.
```
arnefischereit: I do not understand why the link of `ACYCLIC` goes to `REPEATABLE ELEMENTS`, which is quite the opposite.

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,209 @@
...
+A node can be repeated across paths, enabling equijoins to be formed.
```
arnefischereit: Maybe this is clearer as to why we do this:
```suggestion
A node can be repeated across paths, so that path patterns can be connected through xref:patterns/non-linear-patterns/#equijoins[equijoins].
```
We could also mention that this is by virtue of `ACYCLIC` being a path mode. 🤷

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+The actual uniqueness constraint for a given path pattern is the combination of the path mode and the match mode.
+As the default match mode is `DIFFERENT RELATIONSHIPS`, and because the only path mode that can be combined with `REPEATABLE ELEMENTS` is `WALK`, the only path mode that alters the uniqueness constraint is `ACYCLIC`.
```
arnefischereit, line 20: Would this become clearer with a table like https://docs.google.com/document/d/1u7HXEa3RrwZ0JTK4_lLlR1lFeTFkTcNAy7PgHEZ3hyE/edit?tab=t.0#bookmark=id.5dh6dtv8n45o or would that confuse the user even more?

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,209 @@
...
+RETURN p
+----
+
+* *Evaluation:* As the engine traverses the graph, it monitors the nodes already present in the path.
+* *Pruning:* If the engine attempts to move to a node that has already been visited in the current path, it prunes that specific candidate
+solution immediately.
+* *Efficiency:* Because invalid paths are never fully generated or held in memory, execution is faster and uses fewer resources.
+
+
+== Example
```
arnefischereit, line 39: Should this maybe come before the section on performance to establish an understanding of what `ACYCLIC` does?

modules/ROOT/pages/patterns/index.adoc
dogofbrian:
```suggestion
* xref:patterns/path-modes.adoc[] - information about Cypher's different path modes, which determine whether nodes and relationships can appear more than once in a path pattern match.
```
This change might be enough.

modules/ROOT/pages/patterns/path-modes.adoc
dogofbrian: My guess is it would raise more questions.

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,208 @@
...
+Path modes determine whether or not nodes or relationships can appear more than once in a path pattern match.
+Cypher supports three path modes:
+
+* `WALK`: no constraints imposed; both nodes and relationships can be repeated within the path.
+* `TRAIL`: relationships cannot be repeated, but nodes can be repeated within the path.
+* `ACYCLIC`: nodes cannot be repeated within a path. A node can be repeated across paths, enabling equijoins to be formed.
+
+The actual uniqueness constraint for a given path pattern is the combination of the path mode and the match mode.
+As the default match mode is `DIFFERENT RELATIONSHIPS`, and because the only path mode that can be combined with `REPEATABLE ELEMENTS` is `WALK`, the only path mode that alters the uniqueness constraint is `ACYCLIC`.
+This page therefore focuses on the `ACYCLIC` path mode.
```
AlexicaWright, line 21:
```suggestion
This page focuses on the `ACYCLIC` path mode.
```
As I mention below, it reads a bit strange to introduce three path modes and then only talk about one of them. I understand uniqueness constraint here, but perhaps this can be phrased differently. As it is now, it seems as both `WALK` and `TRAIL` are redundant.

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,208 @@
...
+= Path modes
+
+Path modes determine whether or not nodes or relationships can appear more than once in a path pattern match.
+Cypher supports three path modes:
+
+* `WALK`: no constraints imposed; both nodes and relationships can be repeated within the path.
+* `TRAIL`: relationships cannot be repeated, but nodes can be repeated within the path.
+* `ACYCLIC`: nodes cannot be repeated within a path. A node can be repeated across paths, enabling equijoins to be formed.
+
+The actual uniqueness constraint for a given path pattern is the combination of the path mode and the match mode.
```
dogofbrian:
```suggestion
The effective uniqueness constraint for a given path pattern is the combination of the path mode, which operates on path patterns, and the match mode, which operates on the graph pattern level.
A graph pattern is the collection of path patterns in a `MATCH`.
```

edit by rsill-neo4j:
updated the suggestion (removed parentheses, split the sentence, incorporated the suggestion below)

modules/ROOT/pages/patterns/path-modes.adoc
arnefischereit: Or maybe we start with a simpler example just to see how it works like one from the feature tests or [this GH issue](https://github.com/neo4j/neo4j/issues/108) and then come to a real world example like this.

modules/ROOT/pages/patterns/path-modes.adoc
dogofbrian: See my [suggestion](https://github.com/neo4j/docs-cypher/pull/1505/changes#r2989263591)

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,208 @@
...
+As the default match mode is `DIFFERENT RELATIONSHIPS`, and because the only path mode that can be combined with `REPEATABLE ELEMENTS` is `WALK`, the only path mode that alters the uniqueness constraint is `ACYCLIC`.
+This page therefore focuses on the `ACYCLIC` path mode.
+
+For more information on match modes, refer to xref:patterns/match-modes.adoc[].
+
+
+== Syntax
+
+Place the `ACYCLIC` keyword at the head of every path pattern within a MATCH clause.
```
AlexicaWright: Is this different for `WALK` and `TRAIL`?
It reads a bit odd to mention three path modes but exclude two of them.

modules/ROOT/pages/patterns/index.adoc
arnefischereit: Is it path pattern or path pattern match? I have not found a definition for the second one but it seems to me like that is the result of a path pattern, which is the more relevant part: We do not care too much about forbidding `(a)--(a)` but more about `(a)--(b)` not matching a loop.

modules/ROOT/pages/patterns/path-modes.adoc
dogofbrian: Yes, agree.

modules/ROOT/pages/patterns/path-modes.adoc
arnefischereit: I like it. Would this make it clearer?
```suggestion
The actual uniqueness constraint for a given path pattern is the combination of the path mode, which operates on path patterns, and the match mode, which operates at the graph pattern level (a graph pattern is the collection of path patterns in a `MATCH`).
```

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,208 @@
...
+  (e)-[:LINK]->(g), (f)-[:LINK]->(g), (h)-[:LINK]->(i), (h)-[:LINK]->(j),
+  (i)-[:LINK]->(j), (i)-[:LINK]->(z), (g)-[:LINK]->(j), (g)-[:LINK]->(k),
+  (j)-[:LINK]->(k), (j)-[:LINK]->(z), (k)-[:LINK]->(z)
+----
+
+.Router network graph
+image:router-network.png[Router network graph, role=popup]
+
+
+=== TRAIL in packet forwarding
```
AlexicaWright: Using `TRAIL` here sort of defeats the purpose of saying that the page focuses on `ACYCLIC`. I think it should be included, but that the emphasis should be rephrased.

modules/ROOT/pages/patterns/reference.adoc
```diff
@@ -1884,6 +1887,164 @@ RETURN [n IN nodes(p) | n.q] AS nodes
...
+pathKeywords ::= PATH | PATHS
+----
+
+
+=== Rules
+
+==== All path modes in a MATCH must be the same
+
+All path patterns in a single `MATCH` clause must use the same path mode.
+For example, you cannot mix `ACYCLIC` with a pattern that uses the default `WALK` mode:
```
AlexicaWright: The example uses `ACYCLIC` and `TRAIL`.

AlexicaWright (commented):
I'm obviously not an expert on the subject matter, but some thoughts from an editorial perspective.

modules/ROOT/pages/patterns/path-modes.adoc
dogofbrian: Well, we can move the example section up, and start with two trivial queries to contrast default and acyclic path modes.

modules/ROOT/pages/patterns/index.adoc
dogofbrian: Ah yes, that's better. In standards land it's a path binding, but we don't want to go there ...

modules/ROOT/pages/patterns/path-modes.adoc
dogofbrian: Yeah maybe, but the subclauses still seem messy.

What about:

> The scope of a path mode is the path pattern, whereas the scope of a match mode is all of the path pattern under a single `MATCH`. The effective uniqueness constraint at the individual path pattern match is determined by which of the two is more restrictive. As the default match mode is `DIFFERENT RELATIONSHIPS`, and because the only path mode that can be combined with `REPEATABLE ELEMENTS` is `WALK`, the only path mode that alters the uniqueness constraint is `ACYCLIC`. This page therefore focuses on the `ACYCLIC` path mode.

modules/ROOT/pages/patterns/path-modes.adoc
dogofbrian: I tried to lay out the reasons for having the match modes be the way users access either no element uniqueness or relationship uniqueness on the one hand, and path modes for node uniqueness on the other in [this document](https://docs.google.com/document/d/1lAM5fEaLHDCn0MCAs9r3uL0VD9yjonFdL_TcJsLWuUk/edit?usp=sharing). We found that even those working on the feature found the differentiation between path modes and match modes confusing. The reason for there even being match modes was a compromise had to be struck between the different vendors that were working on GQL. The consequence is for Neo4j we have these two path modes that have no effect.

Could we add just an aside to address the question you raised?

modules/ROOT/pages/patterns/path-modes.adoc
rsill-neo4j: the match-modes links were due to copy-paste during editing :) i removed the links

modules/ROOT/pages/patterns/path-modes.adoc
rsill-neo4j: rephrased to account for that

modules/ROOT/pages/patterns/path-modes.adoc
rsill-neo4j: went with the last suggestion, and made it a subsection ("path modes and match modes")

modules/ROOT/pages/patterns/path-modes.adoc
rsill-neo4j: swapped the sections

modules/ROOT/pages/patterns/reference.adoc
rsill-neo4j: rephrased to use TRAIL instead

modules/ROOT/pages/patterns/path-modes.adoc
rsill-neo4j: also, the section heading refers to TRAIL, but there is no TRAIL in the query - is it missing?

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,213 @@
+:description: Information about Cypher's different path modes, which determine whether or not nodes and relationships can appear more than once in a path pattern.
+:page-role: new-neo4j-2026.03 cypher-25-only
+
+= Path modes
+
+Path modes determine whether or not nodes or relationships can appear more than once in a path pattern.
```
arnefischereit: Similar issue as https://github.com/neo4j/docs-cypher/pull/1505#discussion_r2989280848
> Is it path pattern or path pattern match? I have not found a definition for the second one but it seems to me like that is the result of a path pattern, which is the more relevant part: We do not care too much about forbidding (a)--(a) but more about (a)--(b) not matching a loop.


```suggestion
Path modes are levels of constraints that determine whether nodes or relationships can appear more than once in a path pattern match.
```

or `... in a match of a path pattern.`

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+As the default match mode is `DIFFERENT RELATIONSHIPS`, and because the only path mode that can be combined with `REPEATABLE ELEMENTS` is `WALK`, the only path mode that alters the uniqueness constraint is `ACYCLIC`.
+This page therefore focuses on the `ACYCLIC` path mode.
+
+For more information on match modes, refer to xref:patterns/match-modes.adoc[].
+
+
+== Syntax
+
+Place the path mode keyword (`ACYCLIC`) at the head of every path pattern within a MATCH clause.
```
arnefischereit:
```suggestion
Place the path mode keyword (`ACYCLIC`) at the head of every path pattern within a MATCH clause.
The same syntax applies to the other path modes `WALK` and `TRAIL`.
For details, see xref:pattern/reference.adoc#path-modes
```

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,213 @@
...
+Cypher supports three path modes:
+
+* `WALK`: no constraints imposed; both nodes and relationships can be repeated within the path.
+* `TRAIL`: relationships cannot be repeated, but nodes can be repeated within the path.
+* `ACYCLIC`: nodes cannot be repeated within a path. A node can be repeated across paths, enabling equijoins to be formed.
+
+
+== Path modes and match modes
+
+The scope of a path mode is the path pattern, whereas the scope of a match mode is all of the path pattern under a single MATCH.
```
arnefischereit:
```suggestion
The scope of a path mode is the path pattern, whereas the scope of a match mode is all of the path patterns under a single MATCH.
```

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,213 @@
...
+The effective uniqueness constraint at the individual path pattern match is determined by which of the two is more restrictive.
```
arnefischereit: ... while the match mode still applies to the graph pattern match as a whole.

I was certain, this was written somewhere but I could not find it right now and it requires some rephrasing.

modules/ROOT/pages/patterns/path-modes.adoc
arnefischereit: We simply chose to forbid explicitly combining match and path modes, i.e. REPEATABLE ELEMENTS + TRAIL, right?

modules/ROOT/pages/patterns/path-modes.adoc
arnefischereit: `TRAIL` is the effective path mode from the default match mode `DIFFERENT RELATIONSHIPS` and therefore is not typed out. Probably, we should capture this implicit nature here in the title.

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,213 @@
+:description: Information about Cypher's different path modes, which determine whether or not nodes and relationships can appear more than once in a path pattern.
```
arnefischereit:
```suggestion
:description: Information about Cypher's different path modes, which determine whether or not nodes and relationships can appear more than once in a path pattern match.
```

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+[source, cypher, role=test-skip]
+----
+MATCH p = ACYCLIC (start:User \{id: 1})-[:FRIEND]->+(target:User)
+RETURN p
+----
+
+* *Evaluation:* As the engine traverses the graph, it monitors the nodes already present in the path.
+* *Pruning:* If the engine attempts to move to a node that has already been visited in the current path, it prunes that specific candidate
+solution immediately.
```
arnefischereit:
```suggestion
* *Pruning:* If the engine attempts to move to a node that has already been visited in the current path, it prunes that specific candidate from the solution immediately.
```

modules/ROOT/pages/patterns/path-modes.adoc
dogofbrian: Not quite: 

- REPEATABLE ELEMENTS + WALK - allowed
- REPEATABLE ELEMENTS + TRAIL - not allowed
- REPEATABLE ELEMENTS + ACYCLIC - not allowed
- DIFFERENT RELATIONSHIPS + WALK - allowed
- DIFFERENT RELATIONSHIPS + TRAIL - allowed
- DIFFERENT RELATIONSHIPS + ACYCLIC - allowed

But even for the allowed combinations, they are _pointless_ because either the match mode provides the stronger constraint, the match mode and path mode impose the same type of constraint, or the path mode provides the stronger constraint.

modules/ROOT/pages/patterns/path-modes.adoc
dogofbrian: Let's not use the term TRAIL then, as in the example it is actually the default match mode that is enforcing relationship uniqueness

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,214 @@
...
+  (e)-[:LINK]->(g), (f)-[:LINK]->(g), (h)-[:LINK]->(i), (h)-[:LINK]->(j),
+  (i)-[:LINK]->(j), (i)-[:LINK]->(z), (g)-[:LINK]->(j), (g)-[:LINK]->(k),
+  (j)-[:LINK]->(k), (j)-[:LINK]->(z), (k)-[:LINK]->(z)
+----
+
+.Router network graph
+image:router-network.png[Router network graph, role=popup]
+
+
+=== TRAIL in packet forwarding
```
dogofbrian:
```suggestion
=== Relationship uniqueness in packet forwarding
```

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+In many graph contexts, relationship uniqueness (`TRAIL`) is sufficient.
+However, in packet forwarding, a packet should not visit the same router twice, even if it uses a different link.
+A `TRAIL` pattern only ensures that no relationship is reused.
+In the router network graph, the bidirectional links between Router B and Router C, and between Router C and Router D, allow a TRAIL traversal to bounce through the same nodes repeatedly.
```
dogofbrian:
```suggestion
In many graph contexts, the default relationship uniqueness mode of Cypher is sufficient. 
However, in packet forwarding, a packet should not visit the same router twice, even if it uses a different link.
Relationship uniqueness only ensures that no relationship is reused -- it does not prevent nodes from being revisited, meaning cycles are still possible. 
In the router network graph, for example, there are paths where no relationships are repeated but nodes are revisited, such as the cycles involving Router B, C, and D.
```

dogofbrian (approved):
With the suggested changes removing references to TRAIL in the example

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+= Path modes
+
+Path modes are levels of constraints that determine whether nodes or relationships can appear more than once in a path pattern match.
+
+Cypher supports three path modes:
+
+* `WALK`: no constraints imposed; both nodes and relationships can be repeated within the path.
+* `TRAIL`: relationships cannot be repeated, but nodes can be repeated within the path.
+* `ACYCLIC`: nodes cannot be repeated within a path. A node can be repeated across paths, enabling equijoins to be formed.
```
arnefischereit: There was some confusion why no combination of TRAIL (restricting relationships) and ACYCLIC (restricting nodes) is possible: The reason being that restricting nodes implies restricting relationships.

Should we write that in a short sentence here?

dogofbrian:
```suggestion
* `ACYCLIC`: nodes cannot be repeated within a path, which also means relationships cannot be repeated;
a node can, however, be repeated across paths, enabling equijoins to be formed.
```
Something like that?

arnefischereit: That should work, I think.

neo4j-docops-agent:

PR #1514, opened 2026-04-07
Fixes to path modes section

dogofbrian:
Spotted a few issues with the path modes section.


modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -130,8 +130,10 @@ Note that with the `ACYCLIC` path mode, it is not necessary to use predicates to
 1+d|Rows: 1
 |===
 
-You can also use `ACYCLIC` to examine the number of possible routes from the source to the destination router.
-That would be affected by the failure of each:
+You can also use `ACYCLIC` to examine the number of possible routes from the source to the destination router
+that pass through each of the intermediate routers. By grouping the paths on the intermediate nodes and calculating
+the percentage of paths that pass through each router, the sensitivity of the network to a failure at that router
+can be determined:
```
JPryce-Aklundh:
```suggestion
You can also use `ACYCLIC` to examine the number of possible routes from the source to the destination router that pass through each of the intermediate routers.
By grouping the paths based on the intermediate routers and calculating the percentage of total paths that include each router, it is possible to determine the network's sensitivity to a failure at a particular router.
```

JPryce-Aklundh (approved):
Thanks @dogofbrian

PR #1515, opened 2026-04-08
Updates to Path modes docs

JPryce-Aklundh:


modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -35,8 +21,48 @@ RETURN p
...
 For more information, see xref:patterns/reference.adoc#path-modes[Syntax and semantics -> Path modes].
 
+[[path-modes-and-match-modes]]
+== Path modes and match modes
+
+The scope of a path mode is the path pattern, whereas the scope of a xref:patterns/match-modes.adoc[match mode] is all of the path patterns under a single `MATCH`.
+The effective uniqueness constraint at the individual path pattern match is determined by which of the two is more restrictive while the match mode still applies to the graph pattern match as a whole.
 
-== Example
+The table below shows the effective uniqueness constraint for each combination of match mode and path mode.
```
dogofbrian, line 30: The original idea was to skip over each possible combination because it has caused so much confusion just amongst the implementers. We assumed the confusion would be greater for users.

But of course if we can only hide the possible combinations incompletely, which in turn raises questions that the documentation doesn't seem to answer, maybe laying out the options as you do here makes sense.

To compensate, however, I think it would be good to have a bullet point or note box that says something along the lines of:

> in practice what you need to type explicitly is much simpler:
>
> | If you want ... | Then type ... |
> | --- | --- |
> | No uniqueness constraint | MATCH REPEATABLE ELEMENTS ...  |
> | Relationship uniqueness | No keywords |
> | Node uniqueness  | MATCH ACYCLIC ... |
>
> Note: you need to type ACYCLIC at the start of each path pattern

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -70,27 +97,22 @@ In the router network graph, for example, there are paths where no relationships
 
 For example, the following query uses the default behavior for graph element uniqueness, which disallows repeated relationships but allows repeated nodes:
```
dogofbrian: I am wondering if it needs to be made more obvious that this query is just showing that the default uniqueness mode finds paths that packet forwarding would not follow.

modules/ROOT/pages/patterns/path-modes.adoc
```diff
@@ -178,38 +195,37 @@ ORDER BY numPathsWithNode DESC
 
 This shows that router `G` is the weakest point in the network, with 100% of traffic from `A` to `Z` passing through it.
 
-
+[[acyclic-performance-improvement]]
 == Performance improvement with `ACYCLIC`
 
 Using `ACYCLIC` is significantly more efficient than using manual post-filters for node uniqueness.
 
+Prior to the release of the `ACYCLIC` path mode, Cypher would return all paths satisfying the structural requirements of a pattern.
```
dogofbrian:
```suggestion
Prior to the release of the `ACYCLIC` path mode, the default mode of Cypher would return paths satisfying the pattern, including those with nodes repeated.
```

modules/ROOT/pages/patterns/reference.adoc
dogofbrian:
```suggestion
For example, this is not allowed:
```

neo4j-docops-agent:

JPryce-Aklundh:
Closed in favour of: https://github.com/neo4j/docs-cypher/pull/1526

PR #1516, opened 2026-04-09
Add table clarifying vector index capabilities per Neo4j version

JPryce-Aklundh:


mauvo:
Thanks, this is a useful clarification overall. I like the goal of making the versioned capability changes easier to scan.

I think this still needs a few adjustments before merge:

- the new table is duplicated in two places, which creates drift risk
- the wording currently implies that filtering happens via the `WITH` clause, but `WITH` only declares additional properties at index-creation time; the actual filtering happens later in `SEARCH ... WHERE`
-  the reintroduced `[NOTE]` in `indexes/syntax` turns core syntax guidance into admonition content, which makes the section feel warning-heavy rather than explanatory
- if the table stays, the initial version row should be historically precise (`5.11` beta vs `5.13` GA)

I'll make some inline comments too.

I’d be happy with this direction after those points are tightened.

modules/ROOT/pages/indexes/syntax.adoc
```diff
@@ -222,8 +216,30 @@ OPTIONS {
 }
 ----
 
-It is recommended to provide `vector.dimensions` when creating a vector index.
-Doing so ensures that only vectors of that size are indexed and makes dimension mismatches fail clearly at query time.
+[NOTE]
```
mauvo: Could we make this ordinary prose instead of a `[NOTE]`?

My reasoning here is that the recommendation to provide `vector.dimensions` is core syntax guidance, not exceptional or cautionary side information.

Notes work best for content the reader could legitimately treat as an aside. Here, this is part of the main “how to create the index correctly” explanation, so keeping it in normal prose makes the section read more cleanly and consistently.

modules/ROOT/pages/indexes/syntax.adoc
```diff
@@ -222,8 +216,30 @@ OPTIONS {
...
+| *5.13 (initial)*
+| Supports single-label, single-property index for nodes or single-relationship-type, single-property index for relationships.
+Vectors are stored as `LIST <INTEGER \| FLOAT>` properties.
+
+| *2025.10*
+| Adds support for `VECTOR` value properties validated against index dimensions and the configured vector similarity function
+
+| *2026.01*
+| Adds support for multi-label and multi-relationship-type vector indexes with additional properties used for filtering.
+These properties can be used for filtering using the xref:clauses/with.adoc[`WITH`] clause, but they cannot be vector properties.
```
mauvo: I think this wording needs to be clearer.

The additional properties are declared with the `WITH` clause when the index is created, but filtering is not done “using the `WITH` clause”. Filtering happens later in `SEARCH ... WHERE`.

Suggested rewrite:

"These properties can later be used for filtering in `SEARCH ... WHERE`, they cannot be vectors. A vector index can store only one vector property per node/relationship."

modules/ROOT/pages/indexes/syntax.adoc
```diff
@@ -222,8 +216,30 @@ OPTIONS {
...
+
+For a full list of all vector index settings, see xref:indexes/semantic-indexes/vector-indexes.adoc#configuration-settings[Vector index configuration settings].
+
+.Vector index capabilities per Neo4j version
+[cols="a, 3a", options="header"]
+|===
+| Neo4j version
+| Capabilities
+
+| *5.13 (initial)*
```
mauvo: Is `5.13 (initial)` the right milestone here?

My recollection is that vector indexes were introduced in `5.11` as beta and reached GA in `5.13`. If that is right, I think this row should either:
- say `5.11 (beta)`, or
- say `5.13 (general availability)`

modules/ROOT/pages/indexes/syntax.adoc
```diff
@@ -222,8 +216,30 @@ OPTIONS {
...
-It is recommended to provide `vector.dimensions` when creating a vector index.
-Doing so ensures that only vectors of that size are indexed and makes dimension mismatches fail clearly at query time.
+[NOTE]
+It is recommended to provide dimensions when creating a vector index.
+Doing so avoids the confusing situation where the same index may contain side-by-side vectors of different dimensions, while still only allowing comparisons between vectors of the same dimension.
+
+For a full list of all vector index settings, see xref:indexes/semantic-indexes/vector-indexes.adoc#configuration-settings[Vector index configuration settings].
+
+.Vector index capabilities per Neo4j version
+[cols="a, 3a", options="header"]
```
mauvo: I like the table as a presentation device, but repeated in both pages it feels a bit heavy and may drift over time.

Would it make sense to keep the full table only on the canonical `vector-indexes` page, and keep `indexes/syntax` lighter with a short sentence or link?

neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1516.surge.sh

Updated pages:
- [Vector indexes](https://neo4j-docs-cypher-1516.surge.sh/cypher-manual/25/indexes/semantic-indexes/vector-indexes)
- [Syntax](https://neo4j-docs-cypher-1516.surge.sh/cypher-manual/25/indexes/syntax)

neo4j-docops-agent:

PR #1524, opened 2026-04-17
Add vector-3.0 info

JPryce-Aklundh:


modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -469,49 +469,82 @@ DROP INDEX moviePlots
 == Vector index providers for compatibility
 
 It is not possible to specify an index provider when creating indexes in Cypher 25.
-Instead, Neo4j selects the most performant provider (currently `vector-2.0` for vector indexes).
-Previously created `vector-1.0` indexes will continue to function.
+Instead, Neo4j selects the most performant provider (`vector-3.0` for vector indexes as of Neo4j 2026.01).
```
parnmatt:
```suggestion
Instead, Neo4j selects the most performant provider (`vector-3.0` for vector indexes as of Neo4j 2025.09).
```

parnmatt: Unsure if we want to claim "most performant" over "newest" / "latest" … perhaps "most feature rich" if anything.

parnmatt (approved):
It wouldn't hurt to add the introduced neo4j version, such that it is clear that if someone wants the new feature they have to first upgrade, which is different than checking what they currently have.

* `vector-1.0`: 5.11
* `vector-2.0`: 5.18
* `vector-3.0:` 2025.09

Newer providers likely will have the introduced version encoded in the version number,  and thus won't need these extra rows

neo4j-docops-agent:

PR #1538, opened 2026-04-30
Ordering PATH values

JPryce-Aklundh:


modules/ROOT/pages/values-and-types/ordering-equality-comparison.adoc
```diff
@@ -114,6 +114,15 @@ For example, `{b: 100, a: 'foo'}` is less than `{a: 'foo', b: null}` because `['
 
 For more information about `LIST` and `MAP` values, see xref::values-and-types/property-structural-constructed.adoc#constructed-types[Values and types -> Constructed types].
 
+[[ordering-path-values]]
+=== Ordering path values
+
+`PATH` values are compared element-by-element (nodes and relationships alternating); the first differing element determines the order.
+For example, `()-[:A]\->()` is be ordered before `()-[:B]\->()`.
```
gem-neo4j:
```suggestion
For example, `()-[:A]\->()` is ordered before `()-[:B]\->()`.
```

modules/ROOT/pages/values-and-types/ordering-equality-comparison.adoc
```diff
@@ -114,6 +114,15 @@ For example, `{b: 100, a: 'foo'}` is less than `{a: 'foo', b: null}` because `['
...
+
+If one path is a prefix of another, the shorter path is considered smaller (i.e.  it is padded with a special empty value, which is considered less than any other value (including `null`)).
```
gem-neo4j:
```suggestion
If one path is a prefix of another, the shorter path is considered smaller (i.e. it is padded with a special empty value, which is considered less than any other value (including `null`)).
```

gem-neo4j: Do we need the reasoning in the brackets? I feel happy with just knowing it happens, not needing to know how we achieve it

JPryce-Aklundh: No, we probably do not. I agree, brevity is better here. Will fix.

gem-neo4j (approved):
LGTM :)

neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1538.surge.sh

Updated pages:
- [Equality, ordering, and comparison of value types](https://neo4j-docs-cypher-1538.surge.sh/cypher-manual/25/values-and-types/ordering-equality-comparison)

PR #1573, opened 2026-06-08
`DISJOINT BY` documentation

JPryce-Aklundh:


modules/ROOT/pages/subqueries/subqueries-in-transactions.adoc
```diff
@@ -1090,13 +1091,210 @@ The result shows that all transactions are now successful:
...
+
+[[lock-expressions]]
+==== Declaring lock expressions with `DISJOINT BY`
+
+`DISJOINT BY (<expr>, ...)` allows you to declare the expressions identifying locked resources for a `CALL { ... } IN CONCURRENT TRANSACTIONS` statement.
+
+The expressions must use only variables imported from the outer query, since they are evaluated before the subquery runs; for example, `elementId(n)` or `n.id`, where `n` is an imported variable.
+They act as constraints on batch composition: the xref:planning-and-tuning/runtimes/index.adoc[runtime] ensures that no two concurrent batches share a value for any of the declared expressions.
+The first expression acts as the primary grouping key, meaning rows with the same value for it are grouped into the same batch.
+Choosing a low-cardinality first expression therefore leads to better batch grouping, though the order does not affect correctness.
```
henriknyman: This is incorrect. The order of the expressions does not matter. There is no primary grouping, all expressions are equal.

modules/ROOT/pages/subqueries/subqueries-in-transactions.adoc
```diff
@@ -1090,13 +1091,210 @@ The result shows that all transactions are now successful:
...
+
+Batches:
+  Batch A: rows 1, 2  (category=A)
+  Batch B: row  3     (category=A)
+  Batch C: rows 4, 5  (category=B)
+
+Execution:
+  Batch A ------->
+                  Batch B -------->
+  Batch C ------------------------>
```
henriknyman, line 1166: This example is not _exactly_ correct with the current implementation. 
Batch composition is currently greedy in arrival order, not strictly keyed by category, so row 3 and 4 would actually get grouped in the same batch and force fully serial execution. 
This is an implementation detail, and subject to change as we improve the logic (there is an issue in the backlog to investigate different options), so I don't think we need to mention the exact current implementation.

But to make the example truthful with the current implementation it would be enough to add one more input row and still make (almost) the same point:

```
row 1: id=1, category=A
row 2: id=2, category=A
row 3: id=3, category=A
row 4: id=4, category=A
row 5: id=5, category=B
row 6: id=6, category=B

Batches:
  Batch A: rows 1, 2  (category=A)
  Batch B: rows 3, 4  (category=A)
  Batch C: rows 5, 6  (category=B)

Execution:
  Batch A ------->
                  Batch B -------->
  Batch C ------------------------> 
```

modules/ROOT/pages/subqueries/subqueries-in-transactions.adoc
```diff
@@ -1090,13 +1091,210 @@ The result shows that all transactions are now successful:
...
+  MERGE (y:Year {year: line.year})
+  MERGE (m)-[:RELEASED_IN]->(y)
+} IN CONCURRENT TRANSACTIONS
+  DISJOINT BY (line.year, line.movieId)
+  ON ERROR RETRY THEN CONTINUE
+----
+// end::subqueries_in_transactions_deadlock_prevention_disjoint_by_lock_expression[]
+
+`DISJOINT BY (line.year, line.movieId)` declares that `line.year` and `line.movieId` identify the locked resources for each batch.
+`line.year` is the primary grouping key; since many movies share the same release year, it is low-cardinality and leads to efficient batch grouping.
```
henriknyman: Same here, there is no primary grouping key and the ordering is indifferent

modules/ROOT/pages/subqueries/subqueries-in-transactions.adoc
```diff
@@ -1090,13 +1091,210 @@ The result shows that all transactions are now successful:
...
+
+====
+
+
+[[automatic-inference]]
+==== Automatic lock inference with `DISJOINT BY AUTO`
+
+`DISJOINT BY AUTO` instructs the runtime to infer lock expressions automatically from the subquery body, without requiring the user to declare them explicitly.
+This is useful when the locked resources are straightforward to infer statically; for example, `MERGE` patterns on uniquely constrained properties.
+If the runtime cannot infer any lock expressions, the subquery executes with no scheduling restriction.
```
henriknyman, line 1216: We could add a note that when running `EXPLAIN` on the query, the `TransactionApply` or `TransactionForeach` will show the effective disjoint by expressions that were inferred (if there are any) in the Details column. E.g. in the example below `DISJOINT BY (line.person, line.movie)` should appear in the EXPLAIN plan. This can be used to verify if AUTO inference was successful.
(Maybe verify that the order is exactly this with EXPLAIN on the real example)

henriknyman (changes requested):
Nice! 
I had some comments about some incorrect claims, and maybe adding a note or section about how to verify with EXPLAIN if AUTO actually did anything.

henriknyman (approved):
Looks good!

neo4j-docops-agent:

PR #1580, opened 2026-06-16
Clean-up `MERGE` page 

JPryce-Aklundh:


neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1580.surge.sh

Updated pages:
- [Clauses](https://neo4j-docs-cypher-1580.surge.sh/cypher-manual/25/clauses)
- [MERGE](https://neo4j-docs-cypher-1580.surge.sh/cypher-manual/25/clauses/merge)

modules/ROOT/pages/clauses/merge.adoc
```diff
@@ -46,645 +23,457 @@ To recreate the graph, run the following query in an empty Neo4j database:
...
 |===
 
-====
+[[merge-new-patterns]]
+== `MERGE` new patterns
 
-[[merge-merge-single-node-specifying-both-label-and-property]]
-=== Merge single node specifying both label and property
+If `MERGE` cannot match the exact pattern given in the query, it will create the pattern instead.
+To prevent the unintended creation of nodes or relationships that differ in unintended ways from existing ones, it is therefore recommended to use constraints when merging data.
```
rsill-neo4j:
```suggestion
To prevent the creation of nodes or relationships that differ in unintended ways from existing ones, it is therefore recommended to use constraints when merging data.
```
double "unintended"

modules/ROOT/pages/clauses/merge.adoc
```diff
@@ -46,645 +23,457 @@ To recreate the graph, run the following query in an empty Neo4j database:
...
-MATCH
-  (charlie:Person {name: 'Charlie Sheen'}),
-  (oliver:Person {name: 'Oliver Stone'})
-MERGE (charlie)-[r:KNOWS]-(oliver)
-RETURN r
-----
-
-As `Charlie Sheen` and `Oliver Stone` do not know each other in the example graph, this `MERGE` query will create a `KNOWS` relationship between them.
-The direction of the created relationship is left to right.
+Because `Charlie Sheen` and `Michael Douglas` share a chauffeur name (`John Brown`), two separate `Chauffeur` nodes are created.
```
rsill-neo4j:
```suggestion
Even though `Charlie Sheen` and `Michael Douglas` share a chauffeur name (`John Brown`), two separate `Chauffeur` nodes are created.
```

the name being the same is not the reason, is it?
what is the reason then? is there a way to achieve the same with a single clause?

alexfoxgill:
this is well timed, i have some things to add

modules/ROOT/pages/clauses/merge.adoc
```diff
@@ -796,4 +596,106 @@ a| The Cypher planner is able to leverage indexes on property values, however:
...
+== Using `MERGE` with constraints
+
+Constraints on identifying properties (such as `id`) when using `MERGE` prevents duplicating data that differ in unintended ways.
+They also protect against duplicate creation under concurrent loads, where `MERGE` alone only guarantees the existence of the pattern, not its uniqueness.
+
+[role=label--enterprise-edition]
+=== Enforce property existence and uniqueness with key constraints
+
+There are two ways to ensure the uniqueness of a property value in a database: xref:schema/constraints/create-constraints.adoc#create-property-uniqueness-constraints[property uniqueness constraints] and xref:schema/constraints/create-constraints.adoc#create-key-constraints[key constraints].
+Both are backed by indexes, which mean they provide significant performance improvements.
```
rsill-neo4j:
```suggestion
Both are backed by indexes, which provide significant performance improvements.
```

modules/ROOT/pages/clauses/merge.adoc
```diff
@@ -796,4 +596,106 @@ a| The Cypher planner is able to leverage indexes on property values, however:
...
+Key constraints are the preferred option, however, since they both guarantee the existence and uniqueness of a property.
+
+.Create key constraint
+[source, cypher]
+----
+CREATE CONSTRAINT FOR (n:Product) REQUIRE n.id IS NODE KEY
+----
+
+.`MERGE` a node that abides to key constraint
```
rsill-neo4j:
```suggestion
.`MERGE` a node that abides by a key constraint
```

initial reflex, but found this to back it up:
https://jeremybutterfield.com/2020/08/27/abide-by-or-abide-to/

not 100% sure though

alternatives: adheres to, follows, complies with

modules/ROOT/pages/clauses/merge.adoc
```diff
@@ -796,4 +596,106 @@ a| The Cypher planner is able to leverage indexes on property values, however:
...
+
+xref:schema/constraints/create-constraints.adoc#create-property-type-constraints[Property type constraints] are useful to enforce type correctness when merging data.
+
+.Create property type constraint
+[source, cypher]
+----
+CREATE CONSTRAINT FOR (n:Product) REQUIRE n.id IS :: INTEGER
+----
+
+.`MERGE` a node that abides to property type constraint
```
rsill-neo4j:
```suggestion
.`MERGE` a node that abides by a property type constraint
```

(see above)

modules/ROOT/pages/clauses/merge.adoc
```diff
@@ -1,41 +1,18 @@
...
+* If the exact pattern is found, nothing is created; `MERGE` acts like `MATCH` and binds the found pattern.
+* If the exact pattern is not found, `MERGE` acts like `CREATE` and creates the pattern.
 
-Similar to `MATCH`, `MERGE` can match multiple occurrences of a pattern.
-If there are multiple matches, they will all be passed on to later stages of the query.
+`MERGE` also allows for specific actions depending on whether the specified pattern was matched (`ON MATCH`) or created (`ON CREATE`).
 
-The last part of a `MERGE` clause is the `ON CREATE` and/or `ON MATCH` operators.
-These allow a query to express additional changes to the properties of a node or relationship, depending on whether the element was matched (`MATCH`) in the database or if it was created (`CREATE`).
+It is recommended to create constraints prior to merging data: they provide index-backed performance improvements (in the case of property uniqueness and key constraints) while also preventing the creation of data that differ in unintended ways from pre-existing data.
```
rsill-neo4j, line 13: this could link to the relevant section in this page

alexfoxgill:
we have a pre-existing gap in the MERGE docs, specifically around merging relationships under concurrency. for a relationship merge where both end nodes are known, eg:
```cypher
MATCH (a:Person { name: 'alex' })
MATCH (b:Person { name: 'jens' })
MERGE (a)-[:KNOWS]->(b)
```

as expected in serial, executing the query will create the relationship on the first run and then match it on the second run. however under concurrency we **also guarantee** that the `MERGE` will be serialised: we take exclusive locks on both end nodes so that the relationship does not get created twice. 

this behaviour specifically does not use any uniqueness constraints! in fact we cannot express a constraint like "two person nodes may only have a single KNOWS relationship between them", which is probably why we encoded this in `MERGE` behaviour.

there is a [knowledge base article](https://neo4j.com/developer/kb/understanding-how-merge-works/#_merge_acquires_locks_on_nodes_and_relationships_in_the_pattern_when_resulting_in_pattern_creation) referencing this:

> MERGE acquires locks on nodes and relationships in the pattern when resulting in pattern creation
> When MERGE fails to find an existing pattern, it will acquire locks on all bound nodes and relationships in the pattern before creating the missing elements of the pattern.
> 
> This is to ensure a MERGE or CREATE can’t concurrently create the pattern (or alter the properties of an existing pattern to make it identical to the desired pattern) while the MERGE is executing, which would cause pattern duplication.
> 
> After locking on bound elements, MERGE performs another MATCH on the pattern to avoid race conditions where the pattern might get created in the time gap between when MERGE determined the pattern doesn’t exist, and when locks were acquired.

it has come up recently in the context of MVCC where this behaviour is potentially going to be subverted and we discovered that it's not currently explicitly part of the docs

JPryce-Aklundh:
Thanks @alexfoxgill. I have added a section to this end.

modules/ROOT/pages/clauses/merge.adoc
```diff
@@ -296,16 +298,15 @@ RETURN startNode(r) AS startNode,
...
-Also note that the query removes the existing `chauffeurName` properties on any `Person` nodes.
+As the graph contains no `Chauffeur` nodes and no `HAS_CHAUFFEUR` relationships, `MERGE` creates six `Chauffeur` nodes, each with a `name` corresponding to the matched `Person` node's `chauffeurName` property, and a `HAS_CHAUFFEUR` relationship between each pair. 
 
-Because `Charlie Sheen` and `Michael Douglas` share a chauffeur name (`John Brown`), two separate `Chauffeur` nodes are created.
-This differs from the `Location` example above, where `MERGE` bound existing nodes to prevent duplicates.
+This differs from the `Location` example above, where a preceding `MERGE` bound the `Location` nodes before the relationship was created.
+Here, `MERGE` matches on the full pattern including the relationship, so for each `Person` node it finds no existing match and creates a new `Chauffeur` node, even if one with the same name already exists.
```
rsill-neo4j:
```suggestion
Here, `MERGE` matches the full pattern including the relationship, which doesn't match the existing `Person` nodes and instead creates new `Chauffeur` nodes, even if `Person` nodes with the same names already exist.
```

rsill-neo4j (approved):
one more suggestion + approval

alexfoxgill (approved):
nice, sorry i took a while!

neo4j-docops-agent:

PR #1598, opened 2026-07-03
DOCOPS-147 Remove cheat sheet content from PDF

recrwplay:
Adds conditional statements to the cheat sheet index page.

- if the output is PDF, provides a link to the cheat sheet in the published version of the Cypher manual
- if the output is not PDF, outputs the cheat sheet content as normal

Note that the link does not use the `{neo4j-docs-base-uri}` attribute because if it did, the PDF preprocessor would attempt to parse that into a link to the Cypher manual PDF.


neo4j-docops-agent:

JPryce-Aklundh (approved):
Thanks @recrwplay!

PR #1599, opened 2026-07-09
Update vector index documention for the GA of HFQ and the updated default configuration value

neo-walker:


parnmatt:
I do not believe the test failure is due to this docs change, but rather the new index provider being bumped. Though I do find it had to read these test outputs :sweat_smile:

stefano-ottolenghi:
I've made a remark more generic (but equally informative) so we have fewer version numbers to keep up to date, and I've fixed the failing tests. As @parnmatt assumed, it was the provider version having been bumped, while the example results that contained a version had not been updated accordingly. 

> I do find it had to read these test outputs

The assertion error is "Expected property value `vector-2026.07` for key `indexProvider` not found.", and what it's trying to say (before the lengthy dump of results) is: running the query yielded a value `vector-2026.07` for key `indexProvider`, which couldn't be found when scanning through the docs result table.

Does this help demistify the log?

---

On a separate note, in the `vector-2026.07` collapsible, I read:
```
Indexed vector quantization: High-Fidelity Scalar and binary quantization supported.
```
I then did ctrl-f for `High-Fidelity` and found one occurrence only:
```
2026.07: Adds support for High-Fidelity Quantization, which can be enabled [in the index configuration](https://neo4j.com/cypher-manual/25/indexes/semantic-indexes/vector-indexes/#configuration-settings).
```
The linked section however doesn't contain any info on such feature. Is there something else we need to update beside the content of this PR?

stefano-ottolenghi:
TODO: Remember to cherry-pick to cypher 5.

parnmatt:
I'm sure there will be a blog post about it at some point 😅  and @neo-walker will be giving a Nodes talk. It may be worth having a small section about quantisation and the rescoring at some point, along with the other "details you might find interesting". I'm not sure.<br><br>I do remember Jens once noting he'd like to introduce new pages / new manual that sits between cypher and operations manuals where we can go into more interesting details, where things like that could really be expanded. Though I haven't heard if that idea has been pushed more over the last year or so.

stefano-ottolenghi:
Well, as a reader of this page, I am left wondering how I would enable "High-Fidelity Quantization" via configuration settings. Can we link to a specific setting, or clarify how it's enabled? Otherwise we should remove that remark entirely.

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
```diff
@@ -538,9 +538,8 @@ DROP INDEX moviePlots
 [[vector-index-providers]]
 == Vector index providers for compatibility
 
-It is not possible to specify an index provider when creating indexes in Cypher 25.
-Instead, Neo4j selects the most feature-rich provider (`vector-2026.07` for vector indexes as of Neo4j 2026.07).
-Previously created `vector-1.0`, `vector-2.0`, `vector-3.0`, and `vector-2026.06` indexes will continue to function.
+It is not possible to specify an index provider when creating indexes in Cypher 25: Neo4j automatically selects the latest provider.
+Indexes created with older providers will continue to function.
```
parnmatt, line 542: I think that is fine. I think going forward with CalVer provider names, this becomes a bit more clear anyway.

We also made this change so we could bump more often, so it really does make a difference.

It probably wouldn't hurt to ensure that the "latest" is recorded somewhere.

Obviously `vector-2026.07` is the latest as of Neo4j 2026.07 (and hopefully the other indexes will fall instride when we need to update them)

But somewhere noting that the latest is `vector-2026.07` will be useful.
People on versions less than that will at least know that there is a newer version available if they upgrade.
People on newer versions will know that despite being on Neo4j 2026.10 that they are still using the newest and don't need to rebuild.

parnmatt:
it's a combination, setting the quantization to scalar (current default) or binary, and having the expansion factor setting be greater than 1.0 (which it is be default with those quantizations)

modules/ROOT/pages/indexes/semantic-indexes/vector-indexes.adoc
stefano-ottolenghi: > It probably wouldn't hurt to ensure that the "latest" is recorded somewhere.

Oh yeah fair. I've added it beside the collapsible for 2026.07, so we'll just need to remember to move that tag as new versions come.

neo4j-docops-agent:

stefano-ottolenghi:
@neo-walker Is there something else you'd like to add to this, or is it good to merge? :)

neo-walker:
@stefano-ottolenghi I'm happy. Are you doing the cherry pick or should I open another PR?

stefano-ottolenghi:
> @stefano-ottolenghi I'm happy. Are you doing the cherry pick or should I open another PR?

It would be wonderful if you had time to port this to cypher5.

neo-walker:
Cherry pick is up (#1605) and I removed the cherry-pick label from this PR. This is good to merge from my end.

PR #1554, opened 2026-05-21
Update docs for SEARCH operator

pontusmelke:
As of 2026.06 we allow IN as part of the WHERE clause provided with SEARCH

PR #1601, opened 2026-07-13
Document IN filters for vector SEARCH in Neo4j 2026.06

mauvo:
## Summary

- document support for `IN` within vector `SEARCH` filter predicates from Neo4j 2026.06
- add a worked `IN` example, result table, and list-size guidance
- retain the earlier limitation with its version lifted and clarify when list expressions are valid
- link vector index filtering properties directly to the `SEARCH` filtering documentation

This is a targeted backport to the published `cypher-25` branch. The original feature documentation was merged into `dev` in #1554, but the `SEARCH` page changes were not published.

## Validation

- `npm run build:preview`
- `git diff --check`
- live Neo4j 2026.06.0 validation using literal and parameterized lists; both returned the documented results

The live validation used equivalent list-backed embeddings because the Community distribution uses the aligned store format; this does not affect the filter behavior under test.

### DIFF
--- a/modules/ROOT/pages/clauses/search.adoc
+++ b/modules/ROOT/pages/clauses/search.adoc
@@ -478,8 +478,15 @@ RETURN movie.title AS title, movie.releaseDate.year AS year, movie.rating AS rat
 
 [[vector-search-filtering-predicates]]
 === Predicates for vector search with filters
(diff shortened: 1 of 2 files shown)


neo4j-docops-agent:

PR #1602, opened 2026-07-13
SURF-1139: String Interpolation

gem-neo4j:
String interpolation allows users to embed expressions directly within the string to make for a more seamless string building.

e.g `MATCH (p) RETURN s"Hello { p.name }"`


modules/ROOT/pages/values-and-types/boolean-numeric-string.adoc
```diff
@@ -70,3 +70,130 @@ String literals can contain the following escape sequences:
...
+|===
+
+Interpolated strings can be nested, since an embedded expression may itself be an interpolated string.
+
+.Nested string interpolation
+// tag::values_and_types_string_interpolation_nested[]
+[source, cypher]
+----
+WITH 'World' AS name
+RETURN s"Outer: {s'Inner, {name}!'}" AS result
```
rsill-neo4j, line 187: so, this uses the different quotations to make both work in a single string -
what if there are more nesting levels or if you only wanted to use one kind of quotation? would that work with additional WITH statements and variables?

rsill-neo4j (approved):
nice one :)

rsill-neo4j:
i approved already, but here's another thought - we moved the cheat sheet content to this same repo recently, so we can add those parts in the same PR now (`modules/cheat-sheet`)

modules/ROOT/pages/values-and-types/boolean-numeric-string.adoc
gem-neo4j: Ah no the different types of quoting was to show off that it doesn't matter and they can be mixed, would it be better with only using the one type here? 

```
WITH 'World' AS name
RETURN s"Outer: {s"Inner, {name}!"}" AS result
```

Is also be valid

gem-neo4j:
@rsill-neo4j ah I completely forgot the cheat-sheet even existed tbh, I have added a section now :)

neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1602.surge.sh

New pages:
- [<a href="https://neo4j.com/docs/cypher-manual/25/values-and-types/boolean-numeric-string/#string-interpolation" target="_blank" rel="noopener">String interpolation</a>](https://neo4j-docs-cypher-1602.surge.sh/cypher-manual/25/cheat-sheet/string-interpolation)

Updated pages:
- [Additions, deprecations, removals, and compatibility](https://neo4j-docs-cypher-1602.surge.sh/cypher-manual/25/deprecations-additions-removals-compatibility)
- [String concatenation operators](https://neo4j-docs-cypher-1602.surge.sh/cypher-manual/25/expressions/string-operators)
- [Boolean, numeric, and string literals](https://neo4j-docs-cypher-1602.surge.sh/cypher-manual/25/values-and-types/boolean-numeric-string)
- [Cypher Cheat Sheet](https://neo4j-docs-cypher-1602.surge.sh/cypher-manual/25/cheat-sheet)

neo4j-docops-agent:

PR #1603, opened 2026-07-13
Clarify version and limitations for IN vector SEARCH filters

mauvo:
## Summary

- state explicitly that `IN` within vector `SEARCH` filters is supported from Neo4j 2026.06
- retain the earlier limitation with its version lifted
- clarify that lists are valid as the right-hand operand of `IN`, but not with comparison operators
- place the feature label on its own line and improve the supported-predicate wording
- link vector index filtering properties directly to the `SEARCH` filtering documentation

The main `IN` example is already present on `dev` from #1554. This companion keeps the next-release documentation aligned with the targeted current-release backport in #1601.

## Validation

- `npm run build:preview` in a normal repository clone
- `git diff --check`
- live Neo4j 2026.06.0 validation using literal and parameterized lists; both returned the documented results

The live validation used equivalent list-backed embeddings because the Community distribution uses the aligned store format; this does not affect the filter behavior under test.

### DIFF
--- a/modules/ROOT/pages/clauses/search.adoc
+++ b/modules/ROOT/pages/clauses/search.adoc
@@ -479,11 +479,14 @@ RETURN movie.title AS title, movie.releaseDate.year AS year, movie.rating AS rat
 [[vector-search-filtering-predicates]]
 === Predicates for vector search with filters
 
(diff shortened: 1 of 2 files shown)


neo4j-docops-agent:

PR #1604, opened 2026-07-13
SURF-1188: Add example and note about importing WITH and Union branches

### LINKED ISSUE #13842: Deprecated importing WITH in CALL subquery does not propagate into UNION branches, unlike CALL scope clause
**Neo4j Version:** 2026.05.0 Community, also reproduced on 5.26.28 Community  
**Operating System:** Ubuntu 20.04, Linux 5.4.0-216-generic, x86_64  
**Installation Method:** Docker (`neo4j@sha256:6c162e2432f861f2c4e3da77a6ba478e7f10e2160b870541f85294532bc6ff5f`, built 2026-07-02)  
**API:** Cypher via Bolt, Python driver v5.28.4  

### Description

A deprecated `CALL { WITH ... }` subquery behaves differently from the new `CALL (<vars>) { ... }` scope-clause syntax when the subquery contains `UNION` or `UNION ALL`.

In the deprecated form, the variable imported by `WITH a` is visible in the first branch, but is reported as undefined in the second `UNION` branch:

```cypher
MATCH (a:A)
CALL { WITH a
  RETURN a.v AS x
  UNION ALL
  RETURN a.v + 1 AS x
}
RETURN x;
```

This fails with:

```text
CypherSyntaxError: Variable `a` not defined
```

The equivalent new syntax works:

```cypher
MATCH (a:A)
CALL (a) {
  RETURN a.v AS x
  UNION ALL
  RETURN a.v + 1 AS x
}
RETURN x;
```

and returns:

```text
{x: 1}
{x: 2}
```

If this difference is expected because deprecated importing `WITH` is branch-local while `CALL (a)` imports variables into the whole subquery scope, then this is a documentation / migration-clarification issue. Otherwise, it looks like a variable-scope propagation bug in deprecated `CALL { WITH ... }` subqueries containing `UNION`.

### Steps to reproduce

1. Start Neo4j with a fresh database.

2. Create one node:

```cypher
CREATE (a:A {v:1});
```

3. Run a deprecated `CALL { WITH ... }` subquery containing `UNION ALL`, where the second branch references the imported variable:

```cypher
MATCH (a:A)
CALL { WITH a
  RETURN a.v AS x
  UNION ALL
  RETURN a.v + 1 AS x
}
RETURN x;
```

### Expected behavior

One of the following should happen:

1. If `WITH a` imports `a` into the whole deprecated `CALL` subquery, both `UNION ALL` branches should have access to `a`, and the query should return:

```text
{x: 1}
{x: 2}
```

2. If deprecated `CALL { WITH a ... }` only imports `a` into the first `UNION` branch, the behavior should be documented explicitly, especially because the new `CALL (a) { ... }` syntax behaves differently.

### Actual behavior

Neo4j reports:

```text
CypherSyntaxError: Variable `a` not defined
(line 1, column 60)
"MATCH (a:A) CALL { WITH a RETURN a.v AS x UNION ALL RETURN a.v+1 AS x } RETURN x"
                                                           ^
```

The error points to the `a` in the second `UNION ALL` branch:

```cypher
RETURN a.v + 1 AS x
```

This happens even though `a` was imported via `WITH a` at the top of the `CALL` subquery.

The same error occurs with `UNION` and `UNION ALL`.

#### 1. New `CALL (a) { ... }` syntax works

```cypher
MATCH (a:A)
CALL (a) {
  RETURN a.v AS x
  UNION ALL
  RETURN a.v + 1 AS x
}
RETURN x;
```

Result:

```text
{x: 1}
{x: 2}
```

#### 2. Literal instead of variable expression in second branch works

```cypher
MATCH (a:A)
CALL { WITH a
  RETURN a.v AS x
  UNION ALL
  RETURN 99 AS x
}
RETURN x;
```

Result:

```text
{x: 1}
{x: 99}
```

This works because the second branch does not reference `a`.

#### 3. Variable used only in the pre-UNION branch works

```cypher
MATCH (a:A)
CALL { WITH a
  RETURN a.v + 1 AS x
  UNION ALL
  RETURN 99 AS x
}
RETURN x;
```

Result:

```text
{x: 2}
{x: 99}
```

#### 4. `UNION` distinct has the same issue

```cypher
MATCH (a:A)
CALL { WITH a
  RETURN a.v AS x
  UNION
  RETURN a.v + 1 AS x
}
RETURN x;
```

Result:

```text
CypherSyntaxError: Variable `a` not defined
```

#### 5. Importing `a` in each branch works

If the intended deprecated-syntax workaround is to import the outer variable separately in each branch, this form works:

```cypher
MATCH (a:A)
CALL {
  WITH a
  RETURN a.v AS x
  UNION ALL
  WITH a
  RETURN a.v + 1 AS x
}
RETURN x;
```

Result:

```text
{x: 1}
{x: 2}
```

### Additional information

- Reproduced on Neo4j 2026.05.0 Community.
- Also reproduced on Neo4j 5.26.28 Community.
- Not a recent 2026.05.0 regression.
- Affects both `UNION` and `UNION ALL`.
- The new `CALL (a) { ... }` scope-clause syntax works correctly.
- The deprecated `CALL { WITH a ... }` form appears to import `a` only into the first `UNION` branch, not the whole subquery body.

### Why this matters

The old and new subquery import forms look like they should be migration-equivalent:

```cypher
CALL {
  WITH a
  ...
}
```

versus:

```cypher
CALL (a) {
  ...
}
```

However, in the presence of `UNION` / `UNION ALL`, they have different variable visibility behavior.

This can break migrated or legacy queries in a non-obvious way. The failure is especially confusing because:

- the variable is imported at the top of the `CALL` subquery;
- the first branch can use it;
- the second branch cannot;
- the new `CALL (a)` syntax can use it in both branches.

### Requested clarification

Please clarify one of the following:

1. If deprecated `CALL { WITH a ... }` is intended to import `a` into the whole subquery, then the second `UNION` branch should also be able to reference `a`.

2. If deprecated importing `WITH` is intended to be branch-local in `UNION` subqueries, please document this explicitly and recommend either:

```cypher
CALL (a) {
  ...
}
```

or:

```cypher
CALL {
  WITH a
  ...
  UNION ALL
  WITH a
  ...
}
```

as the correct form.

gem-neo4j:
Fixes: https://github.com/neo4j/neo4j/issues/13842


modules/ROOT/pages/subqueries/call-subquery.adoc
```diff
@@ -368,6 +368,7 @@ To return imported variables they must be renamed.
 
 * The importing `WITH` clause must the first clause of a subquery (or the second clause, if directly following a `USE` clause).
```
rsill-neo4j:
```suggestion
* The importing `WITH` clause must be the first clause of a subquery (or the second clause, if directly following a `USE` clause).
```

modules/ROOT/pages/subqueries/call-subquery.adoc
```diff
@@ -368,6 +368,7 @@ To return imported variables they must be renamed.
...
 * It is not possible to follow an importing `WITH` clause with any of the following clauses: `DISTINCT`, `ORDER BY`, `WHERE`, `SKIP`, and `LIMIT`.
+* When using a `UNION [ALL]` directly within the `CALL` any imported variables need to be imported per `UNION` branch.
```
rsill-neo4j:
```suggestion
* When using a `UNION [ALL]` directly within the `CALL` any imported variables must be imported per `UNION` branch.
```

modules/ROOT/pages/subqueries/call-subquery.adoc
```diff
@@ -421,6 +422,60 @@ RETURN largeLists
 
 |===
 
+Attempting to use an outer variable only imported into the first `UNION` branch in the second will throw an error:
```
rsill-neo4j:
```suggestion
Attempting to use an outer variable only imported into the first `UNION` branch in the second throws an error:
```

rsill-neo4j (approved):
Minor suggestions + approval

rsill-neo4j:
this is also a general fix, right?

gem-neo4j:
@rsill-neo4j yes :)

neo4j-docops-agent:

PR #1607, opened 2026-07-14
document GROUP BY

rsill-neo4j:


modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,483 @@
+:description: `GROUP BY` groups results by a grouping element or a list of grouping elements.
+:table-caption!:
+:page-role: new-neo4j-2026.07 cypher-25-only
+
+[[query-group-by]]
+= GROUP BY
+
+`GROUP BY` is a subclause that allows you to group the results of a `RETURN` or `WITH` clause.
```
JoelBergstrand:
```suggestion
`GROUP BY` is a subclause that allows you explicit control over the grouping keys of a `RETURN` or `WITH` clause.
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,483 @@
...
+
+`GROUP BY` arranges identically structured data into groups based on one or more columns.
```
JoelBergstrand:
```suggestion
`GROUP BY` arranges identically structured data into groups based on one or more grouping keys.
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,483 @@
...
+Cypher's `GROUP BY` functions similarly to SQL's `GROUP BY` and it is also xref:appendix/gql-conformance/supported-optional.adoc#gq15[GQL-compliant].
+
+
+== Grouping keys
+
+The following rules apply to grouping keys:
+
+* A single grouping key or multiple grouping keys are possible
+* The set of grouping keys may also be empty
+* `GROUP BY ALL` makes every return item a grouping key
```
JoelBergstrand:
```suggestion
* `GROUP BY ALL` makes every non-aggregating return item a grouping key
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,483 @@
...
+| 2016 | 4.5    
+| 2019 | 3.6    
+2+d|Rows: 7
+|===
+
+`title` is not returned, but still used as a grouping key.
+=====
+
+
+== Variable visibility
```
JoelBergstrand: I'm not convinced this section is useful, or could be a line in the rules.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,483 @@
...
+| actor_name         | genre    | year | averageRating 
+
+| "Jessica Chastain" | "Drama"  | 2016 | 4.5           
+| "Jessica Chastain" | "Drama"  | 2017 | 4.4           
+| "Jessica Chastain" | "Action" | 2019 | 3.5           
+| "Jessica Chastain" | "Horror" | 2019 | 3.6           
+| "Jessica Chastain" | "Drama"  | 2022 | 4.0           
+| "James McAvoy"     | "Horror" | 2019 | 3.6           
+| "Eddie Redmayne"   | "Drama"  | 2022 | 4.2           
+3+d|Rows: 5
```
rsill-neo4j:
```suggestion
3+d|Rows: 7
```

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: commented out for now

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -55,6 +55,23 @@ OPTIONS {
...
 
+=== New
+
+[cols="2", options="header"]
+|===
+| Feature
+| Details
+
+a|
+label:functionality[]
```
rsill-neo4j:
```suggestion
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
+:description: `GROUP BY` groups results by a grouping key or a list of grouping keys.
+:table-caption!:
+:page-role: new-neo4j-2026.07 cypher-25-only
+
+[[query-group-by]]
+= GROUP BY
+
+`GROUP BY` is a subclause that allows you explicit control over the grouping keys of a `RETURN` or `WITH` clause.
+Cypher's implicit grouping stays the same.
```
hvub: You may want to link here to https://neo4j-docs-cypher-1607.surge.sh/cypher-manual/25/functions/aggregating/#grouping-keys

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+
+== Grouping keys
+
+The following rules apply to grouping keys:
+
+* A single grouping key or multiple grouping keys are possible
+* The set of grouping keys may also be empty
+* `GROUP BY ALL` makes every non-aggregating return item a grouping key
+* If a projection clause contains a `GROUP BY` clause, then every projection item must be:
+** A leaf, that is an aggregation function, constant, parameter, local variable or a grouping key expression
```
hvub: I am not sure local variable is correct here.

Say you have:

```
MATCH (a)-->(b)-->(c)
RETURN COUNT(c), 1, $p, b, a
 GROUP BY a
```

- The return item `COUNT(c)` is "aggregation function", i.e. allowed by the first kind of leaf
- The return item `1` is "constant", i.e. allowed by the second kind of leaf
- The return item `$p` is "parameter", i.e. allowed by the third kind of leaf
- The return item `b` is "local variable", I think, i.e. allowed by the forth kind of leaf
- The return item `a` is "a grouping key expression", i.e. allowed by the fifth kind of leaf

However, the `b` is invalid.

I guess it comes down to what "local variable" precisely means. Subexpression may introduce variables that are local to the subexpression and these would be valid. Example:

```
MATCH (a)-->(b)-->(c)
RETURN COUNT(c), 1, $p, [b IN a.prop | b+1], a
 GROUP BY a
```

Here, the first `b` in `[b IN a.prop | b+1]` is the declaration of a `b` variable local to the list comprehension. It shadows the `b` from the `MATCH` in the list comprehension (in the `WHERE ... | ...` part of the list comprehension to be precise.). The second `b` in `[b IN a.prop | b+1]` references this local `b` (and not the `b` from the `MATCH`). This is a valid reference.

I guess that is what is behind "local variable". In that sense "local variable" correct and the reply to my example with just return item `b` above would be. "But `b` is not a local variable here."

Fine. But I claim that subtlety is lost to 80% of the readers. Maybe we called it "expression-local variable" to stress a bit more what "local" refers to.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+** An exact grouping key, equal to a grouping key expression
+* Grouping keys which are not also return items are not projected to the output
+
+
+[[example-graph]]
+== Example graph
+
+The following graph is used for the examples below:
+
+image::graph-group-by-clause.svg[Example graph for GROUP BY examples,role=popup]
```
hvub: The image in the preview on show one Person node, while the Cypher below has three.

Bonus observation: midsommar has no relationship to any Person. That is not a problem as such, but may be unintended. That is why I am mentioning it.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+Cypher's `GROUP BY` functions similarly to SQL's `GROUP BY` and it is also xref:appendix/gql-conformance/supported-optional.adoc#gq15[GQL-compliant].
+
+
+== Grouping keys
+
+The following rules apply to grouping keys:
+
+* A single grouping key or multiple grouping keys are possible
+* The set of grouping keys may also be empty
+* `GROUP BY ALL` makes every non-aggregating return item a grouping key
```
hvub, line 23: You could add here and item for `GROUP BY ()`, e.g.

```suggestion
* `GROUP BY ()` groups with no grouping key, i.e. all rows from one big group
* `GROUP BY ALL` makes every non-aggregating return item a grouping key
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+|===
+
+`GROUP BY ALL` uses all projected expressions as grouping keys.
+In this case, only the two drama movies from 2022 starring Jessica Chastain actually calculate an average from two values.
+=====
+
+
+== GROUP BY and ORDER BY
+
+[NOTE]
```
hvub: The note comes before example, so you kind of expect the example to be about the shadowing the note talks about, but it isn't. Maybe move the note after the example. Also, the example demos the general/normal case, while the note talks about a special case. It is typically good to go from general/normal to special and not vice versa. Push the messy details to the back, so to speak.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+| 2019 | 3.5    
+| 2019 | 3.6    
+| 2022 | 3.8    
+| 2022 | 4.2    
+| 2016 | 4.5    
+    
+2+d|Rows: 6
+|===
+
+You can alias expressions with `LET` and use the alias in the `RETURN` (or `WITH`) clause as well as the `GROUP BY` clause.
```
hvub: This points out `LET`, which is not wrong. However, to `RETURN ... GROUP BY ...`, the `year` is just a regular variable introduce by some clause before it. I would not even call `year` an alias. 

But more interesting part here is that `GROUP BY` also refers to an alias introduce be the RETURN, namely `rating`.

The term "alias" is strongly link to names introduce by `AS <name>` in the close proximity of that typically  limited to the same clause (`RETURN ... ` in this example) and its subclauses (`GROUP BY ...` in this example). For subsequent clauses, it just a variable or result column as every other variable or result column. 

So, I would not mention `LET` but point to `m.rating as rating` and the use of `rating` in `GROUP BY`.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+| 2019 | 3.5    
+| 2019 | 3.6    
+| 2022 | 3.8    
+| 2022 | 4.2    
+| 2016 | 4.5    
+| 2019 | 3.6    
+2+d|Rows: 7
+|===
+
+`title` is not returned, but still used as a grouping key.
```
hvub, line 340: This example is a bit unfortunate in that it does not show the effect in the result, title does not make the grouping more fine-grained. 

Consider 

```
MATCH (m: Movie)
LET title = m.title
RETURN m.year as year
  GROUP BY title, year
```

as an alternative.

If you wonder, what the point of this functionality is. It is when you want to use that grouping key in an aggregating expression, but not project it itself. 

```
MATCH (m: Movie)
LET title = m.title
RETURN m.year as year, AVG(m.rating) * size(title) as funnyRating
  GROUP BY title, year
```

This is obviously made up in the constraints of the example data. More realistic use cases are when you have `MATCH (a)-->(b)` where it is known from the business domain that every `a` has just one `b`, so reporting something about group of `a` and `b` can be labeled in a result just by `a`, so that the user does not want `b` in the result. But if `b` is needed somewhere else outside an aggregate function is still has to be a grouping key. With GROUP BY you can make it a grouping key without it showing up as a result column.

To achieve the same effect without GROUP BY you would have to do this:

```
MATCH (m: Movie)
LET title = m.title
WITH m.year as year, title, AVG(m.rating) * size(title) as funnyRating
RETURN title, funnyRating
```

I leave to you to which degree you want to use this input to change or extend the example here.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
+:description: `GROUP BY` groups results by a grouping key or a list of grouping keys.
+:table-caption!:
+:page-role: new-neo4j-2026.07 cypher-25-only
+
+[[query-group-by]]
+= GROUP BY
+
+`GROUP BY` is a subclause that allows you explicit control over the grouping keys of a `RETURN` or `WITH` clause.
```
gem-neo4j:
```suggestion
`GROUP BY` is a subclause that gives explicit control over the grouping keys of a `RETURN` or `WITH` clause.
```

modules/ROOT/pages/clauses/group-by.adoc
gem-neo4j: Can you crop the image, it is really zoomed out

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+| genre    | averageRating 
+
+| "Drama"  | 4.225         
+| "Action" | 3.5           
+| "Horror" | 3.6           
+        
+2+d|Rows: 3
+|===
+
+The four drama movies have an average rating of `4.225`.
```
gem-neo4j, line 116: This is a bit strange as it only talks about one of the 3 returned results, perhaps say we get to see the average rating for each genre e.g drama movies have an average rating of `4.225`?

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+`GROUP BY ALL` uses all projected expressions as grouping keys.
+In this case, only the two drama movies from 2022 starring Jessica Chastain actually calculate an average from two values.
+=====
+
+
+== GROUP BY and ORDER BY
+
+[NOTE]
+====
+Neo4j 2026.07 features changes to the visibility of expressions in `WHERE` and `ORDER BY` clauses when using `GROUP BY`:
```
gem-neo4j:
```suggestion
Neo4j 2026.07 features changes the visibility of expressions in `WHERE` and `ORDER BY` clauses when using `GROUP BY`:
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+
+Projected aliases (variables named with `AS` in `RETURN` and `WITH` clauses) shadow variables defined prior to `RETURN` or `WITH`, if they share the same name, in the scope of subclauses (like `WHERE` and `ORDER BY`).
```
gem-neo4j: Can we explain this better? 😅

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+| "Jessica Chastain" | "Action" | 3.5           
+| "Jessica Chastain" | "Drama"  | 4.225         
+| "Eddie Redmayne"   | "Drama"  | 4.2           
+| "Jessica Chastain" | "Horror" | 3.6           
+| "James McAvoy"     | "Horror" | 3.6             
+        
+3+d|Rows: 5
+|===
+
+Grouping by actor and genre, the results can also be ordered by genre.
```
gem-neo4j:
```suggestion
The results are ordered by genre after being grouped.
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+| 2017 | 4.4    
+| 2019 | 3.5    
+| 2019 | 3.6    
+| 2022 | 3.8    
+| 2022 | 4.2    
+| 2016 | 4.5    
+2+d|Rows: 6
+|===
+
+In this query, `GROUP BY` is equivalent to xref:clauses/return.adoc#return-unique-results[`DISTINCT`].
```
gem-neo4j:
```suggestion
In this query, `GROUP BY` is equivalent to using the xref:clauses/return.adoc#return-unique-results[`DISTINCT`] keyword.
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+
+.GQLSTATUS error chain
+[role="queryresult",stripes="all", cols="1", ]
+|===
+| link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42I18/[42I18]: error: syntax error or access rule violation - reference to non-grouping sub-expression. The expression contains a non-grouping sub-expression `m`. In an aggregating context only grouping sub-expressions and constants are allowed.
+
+link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42001/[42001]: error: syntax error or access rule violation - invalid syntax
+|===
+
+42I18 is technically not specific to `GROUP BY` but rather to grouping keys.
```
gem-neo4j: This feels irrelevant

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+For another example, see the error in example 28 in xref:functions/aggregating.adoc#grouping-key-examples[Aggregation expressions and grouping keys -> Examples].
```
gem-neo4j:
```suggestion
For more examples of aggregation errors, see xref:functions/aggregating.adoc#grouping-key-examples[Aggregation expressions and grouping keys -> Examples].
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+
+.GQLSTATUS error chain
+[role="queryresult",stripes="all", cols="1", ]
+|===
+| link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42N44/[42N44]: error: syntax error or access rule violation - inaccessible variable. It is not possible to access the variable `m` declared before the RETURN clause when using `DISTINCT`, an aggregation, or a `GROUP BY` clause.
+
+link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42001/[42001]: error: syntax error or access rule violation - invalid syntax
+|===
+
+42N44 is technically not specific to `GROUP BY`.
```
gem-neo4j: Same as before, doesn't feel like useful info

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+=====
+
+[source, cypher]
+---- 
+MATCH (m)
+RETURN m as a
+GROUP BY a.year
+----
+
+// TODO: GQL error code
```
gem-neo4j: Todo

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+.GQLSTATUS error chain
+[role="queryresult",stripes="all", cols="1", ]
+|===
+| link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42I80/[42I80]: error: syntax error or access rule violation - invalid grouping element. The grouping element 'a.year' is not a valid grouping key. A grouping element that references the projection item alias `a` must be a simple variable reference.
+
+link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42001/[42001]: error: syntax error or access rule violation - invalid syntax
+|===
```
gem-neo4j, line 413: Maybe give a short sentence explanation like the other examples?

modules/ROOT/pages/functions/aggregating.adoc
```diff
@@ -1169,13 +1170,15 @@ RETURN p.age, p.age - max(f.age)
...
+Cypher only recognizes a sub-expression in aggregating expressions as a grouping key if the grouping key is:
 
 - A variable - e.g. the `p` in `RETURN p, p.age - max(f.age)`.
 - A property access - e.g. the `p.age` in `RETURN p.age, p.age - max(f.age)`.
 - A map access - e.g. the `p.age` in `WITH {name:'Keanu Reeves', age:58} AS p RETURN p.age, p.age - max(p.age)`.
 
-If more complex grouping keys are needed as operands in aggregating expression, it is always possible to project them in advance using `WITH`.
+These are called _recognizable grouping key expressions_.
+
+If you need more complex grouping keys as operands in aggregating expressions, you can always project them with `AS` in `RETURN` and `WITH`.
```
gem-neo4j:
```suggestion
If you need more complex grouping keys as operands in aggregating expressions, you can project them with `AS` in `RETURN` and `WITH`.
```

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -55,6 +76,25 @@ OPTIONS {
...
+
+a|
+label:new[]
+[source, cypher, role="noheader"]
+----
+MATCH (p:Person)-[:ACTED_IN]->(m:Movie)
+RETURN p.name AS actor_name, m.genre AS genre, avg(m.rating) AS averageRating
+GROUP BY actor_name, genre
+----
+| A new xref:clauses/group-by.adoc[GROUP BY] subclause for `WITH` and `RETURN` has been added which lets user explicitly state the grouping keys for the query.
```
gem-neo4j:
```suggestion
| A new xref:clauses/group-by.adoc[GROUP BY] subclause used in conjunction with `WITH` and `RETURN` clauses has been added. It allows a user to explicitly state the grouping keys for the query.
```

gem-neo4j (commented):
Who is in charge of the script which makes the keywords highlighted in Cypher queries? Seems quite a few of our newer keywords are missing e.g LET ad now GROUP BY

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+----
+
+.GQLSTATUS error chain
+[role="queryresult",stripes="all", cols="1", ]
+|===
+| link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42I18/[42I18]: error: syntax error or access rule violation - reference to non-grouping sub-expression. The expression contains a non-grouping sub-expression `m`. In an aggregating context only grouping sub-expressions and constants are allowed.
+
+link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42001/[42001]: error: syntax error or access rule violation - invalid syntax
+|===
+
```
hvub, line 366: It may be worth explaining here that the error points out that `m` in return item `m.rating AS rating` is not a grouping key — not listed in GROUP BY — and hence can't be user outside an aggregate function.

Possible changes to turn this into valid query, depending on what the user wants to accomplish are:

Do not project `m.rating AS rating`:
```
MATCH (m:Movie)
LET title = m.title
RETURN m.rating AS rating, avg(m.rating) AS avg_rating
GROUP BY title
```

Project `title` instead:
```
MATCH (m:Movie)
LET title = m.title
RETURN title, m.rating AS rating, avg(m.rating) AS avg_rating
GROUP BY title
```

Make `m` a grouping key (unlikely what the user wants, since `m` is also aggregated)
```
MATCH (m:Movie)
LET title = m.title
RETURN title, m.rating AS rating, avg(m.rating) AS avg_rating
GROUP BY title, m
```

I leave it to you to decide if this fix should be part to of the page. I list them just FYI.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+----
+
+.GQLSTATUS error chain
+[role="queryresult",stripes="all", cols="1", ]
+|===
+| link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42N44/[42N44]: error: syntax error or access rule violation - inaccessible variable. It is not possible to access the variable `m` declared before the RETURN clause when using `DISTINCT`, an aggregation, or a `GROUP BY` clause.
+
+link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42001/[42001]: error: syntax error or access rule violation - invalid syntax
+|===
+
```
hvub, line 390: Here an explanation that goes a little a beyond what the error message is stating would be:

Variable `m` cannot be referenced in `ORDER BY` when `RETURN` (or `WITH`) are aggregating, since `m` is not a grouping key, i.e. gone after the grouping and aggregation, and `ORDER BY` happens after grouping and aggregation.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,387 @@
...
+// TODO: GQL error code
+.GQLSTATUS error chain
+[role="queryresult",stripes="all", cols="1", ]
+|===
+| link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42I80/[42I80]: error: syntax error or access rule violation - invalid grouping element. The grouping element 'a.year' is not a valid grouping key. A grouping element that references the projection item alias `a` must be a simple variable reference.
+
+link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42001/[42001]: error: syntax error or access rule violation - invalid syntax
+|===
+
```
hvub, line 414: Here the error message says it pretty well. It may be still worth repeating in normal text, since there error message test is rendered pretty small and does not feel like part of the main narrating view when you read the page.

(I can given more color on the reasoning behind the error, but that is less fundamental so less vital for a user's understanding. Hence, only upon request.)

modules/ROOT/pages/functions/aggregating.adoc
hvub, line 1: Would it be useful if this paper would also point to `GROUP BY` in some way? Like say in some fitting place something like "For more precise control of the grouping keys the use of the GROUP BY clause is recommended" and/or in

```
*Grouping keys* are non-aggregating expressions that are used to group the values going into the aggregating functions.
For example, given the following query containing two return expressions, `n` and `+count(*)+`:
```

add

```
*Grouping keys* either given implicit or explicitly.
Implicit grouping keys are non-aggregating expressions that are used to group the values going into the aggregating functions.
For example, given the following query containing two return expressions, `n` and `+count(*)+`:
Explicit grouping keys are given by the `GROUP BY` clause [link].
```

Something like that. A user who looks up aggregation should here about `GROUP BY`.

rsill-neo4j:
> Who is in charge of the script which makes the keywords highlighted in Cypher queries? Seems quite a few of our newer keywords are missing e.g LET ad now GROUP BY

it's not only surge, but also the live docs :thinking: 
@recrwplay is that something we control?

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j:
```suggestion
```

recrwplay:
> > Who is in charge of the script which makes the keywords highlighted in Cypher queries? Seems quite a few of our newer keywords are missing e.g LET ad now GROUP BY
> 
> it's not only surge, but also the live docs 🤔 @recrwplay is that something we control?

@rsill-neo4j we use a highlight package, and for cypher we require [highlightjs-cypher](https://github.com/highlightjs/highlightjs-cypher). If these are new keywords, they are probably missing from that be missing from that.

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: i split the sentence and tried to make it clearer :+1:

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -27,6 +27,27 @@ For more information, see xref:queries/select-version.adoc[].
...
+label:deprecated[]
+[source, cypher, role="noheader"]
+----
+RETURN a.p + 0 AS b, COUNT(*) AS c 
+ORDER BY a.p + 0 + 1
+----
+a| Deprecated the following:
+
+- subclause expressions which reference projection item expressions which are more complex than a variable or a property access on a variable.
+- cases with an ambiguity between an incoming variable and a newly declared variable in the subclause expression.
```
hvub: cf. comment about the term "alias". "newly declared variable" is not incorrect here, but I would find it more fitting to say "new alias" or just "alias".

```suggestion
- cases with an ambiguity between an incoming variable and a alias in the subclause expression.
```

Worth pointing out that the example only illustrates the first item ("subclause expressions which reference projection item expressions which are more complex than a variable or a property access on a variable.").

An example for the second item would be

```
MATCH (a)-->(b)
RETURN a.p AS a, COUNT(*) AS c 
ORDER BY a.p
```

or

```
MATCH (a)-->(b)
RETURN a.p AS x, b AS a, COUNT(*) AS c 
ORDER BY a.p
```

In both cases, it is unclear if the `a` in ORDER BY refers to the alias `a` in return item `b AS a` or the variable `a` from `MATCH (a)-->(b)`.

GQL gives this clear semantics (the alias wins). Our existing implementation (without `GROUP BY`) has inconsistent handling of this. To make it consistent with GQL we have break queries. This deprecation is meant to remove these cases from the user base over time, so that in some glorious future we can given it the GQL semantics.

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -55,6 +76,25 @@ OPTIONS {
...
+a|
+label:new[]
+[source, cypher, role="noheader"]
+----
+MATCH (p:Person)-[:ACTED_IN]->(m:Movie)
+RETURN p.name AS actor_name, m.genre AS genre, avg(m.rating) AS averageRating
+GROUP BY actor_name, genre
+----
+| A new xref:clauses/group-by.adoc[GROUP BY] subclause for `WITH` and `RETURN` has been added which lets user explicitly state the grouping keys for the query.
+This is an alternative to the implicit grouping available in Cypher.
```
hvub, line 133: Any point in mentioning here that GROUP BY is GQL conformant? Have we done this for similar case (e.g. LET, NEXT, FINISH)?

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
```diff
@@ -283,6 +283,11 @@ For example, GQL’s graph reference values `CURRENT_GRAPH` and `CURRENT_PROPERT
 | xref:clauses/order-by.adoc[`ORDER BY`]
 |
 
+| [[gq15]] GQ15
+| `GROUP BY` clause
+| xref:clauses/group-by.adoc[`GROUP BY`]
+| Cypher's version of `GROUP BY` currently doesn't fully implement the GQL version: GQL requires sort keys to be able to reference all elements in the `GROUP BY` clause and all return item aliases. This conflicts with the current Cypher behavior as aliases are not evaluated in Cypher until after sorting.
```
hvub: This sound wrong to me. Particularly the second sentence. There is this coming coming from?

From the description I would think this query would not work and fail with some syntax error:
```
MATCH (a)-->(b)-->(c)
RETURN a AS x, b AS y, COUNT(*) AS d 
  GROUP BY x, y, c
  ORDER BY id(a), x.p, id(y), b.p, c, id(c)
```
But is work on dev. (Maybe it is different on the release branch, I have not checked that.)

So I think this should be:

```suggestion
| 
```

Or what am I missing?

hvub (commented):
Quite a number of comment but nothing to earth shattering. Fine tuning. Overall this looks good.

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: uploaded a new PNG file inbetween reviews :)
yes, i added midsommar because i was planning to do more with the graph but then didn't have to
it's still used in some of the examples and contributes to eg the averageRating of horror movies. neat observation though :)

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: added examples to all of the cases and made it a list of items for better readability

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: added `GROUP BY ()` in parentheses

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: but this then isn't a special case anymore, at least not with how the page is designed as a whole currently (because the "basic examples" use `AS`-aliases)

i rephrased the explanation:
"You can group by aliased and unaliased return items (`year` is not aliased, `rating` is aliased)."

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: it does make it more refined, at least when i tested it:

```
neo4j@neo4j> MATCH (m: Movie)
             LET title = m.title
             RETURN m.year AS year, m.rating AS rating
               GROUP BY title, year, rating;
+---------------+
| year | rating |
+---------------+
| 2016 | 4.5    |
| 2017 | 4.4    |
| 2019 | 3.5    |
| 2019 | 3.6    |
| 2019 | 3.6    |
| 2022 | 3.8    |
| 2022 | 4.2    |
+---------------+

7 rows
ready to start consuming query after 25 ms, results consumed after another 2 ms
neo4j@neo4j> MATCH (m: Movie)
             LET title = m.title
             RETURN m.year AS year, m.rating AS rating
               GROUP BY year, rating;
+---------------+
| year | rating |
+---------------+
| 2016 | 4.5    |
| 2017 | 4.4    |
| 2019 | 3.5    |
| 2019 | 3.6    |
| 2022 | 3.8    |
| 2022 | 4.2    |
+---------------+

6 rows
```

one more row with `title`

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: added some error explanations in the meantime (i think Gem's feedback was along those lines as well).

thanks for the suggestions, i'll update what i came up with (which wasn't entirely correct either)

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
rsill-neo4j: added the example and accepted the suggestion

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
rsill-neo4j: afaik, we do not point out GQL conformance in the additions, ... page
it's in the GROUP BY page, and the GQL conformance page also lists it ofc

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
rsill-neo4j: Appendix of CIP-236 (https://docs.google.com/document/d/1Xf7Kh6QJfWx3jglC5TbkN8zg6s58QlGU0XCXShwoB-U/edit?pli=1&tab=t.0):
"Current issue - Reconciling GQL compliance with Cypher"

is this wrong/outdated?

modules/ROOT/pages/clauses/group-by.adoc
hvub: Figure looks way better, particularly with the cropping Gem mentioned. The Cypher CREATE below and the figure are still not aligned though. midsommar is in the Cypher but not in the figure. To make it consistent, you can:

1. remove midsommar from the CREATE (probably breaks some of the examples)
2. add midsommar to the figure
a. if you care that it should not be isolate it, then connect with a relationship it to something (obviously the relationship should also show up in the figure)
b. if you do not care out that, show it isolated in the figure

Either is fine by me. Main point is that figure and CREATE should be consistent.

modules/ROOT/pages/clauses/group-by.adoc
hvub: Ah. Okay. I checked to duplicate lines but misses `| 2019 | 3.6 |` is a duplicate. Sorry. Then the example is good. Maybe ordering by year and rating would help to make it standout a bit more. Or even some text explicitly pointing out the duplicate in the result caused by not projecting one of the grouping keys.

modules/ROOT/pages/functions/aggregating.adoc
```diff
@@ -1055,6 +1055,8 @@ For instance, `0.1 * (sum(x.a) / count(x.b))` is an aggregating expression that
 Both are input arguments to the division expression.
 
 *Grouping keys* are non-aggregating expressions that are used to group the values going into the aggregating functions.
+Grouping keys are implicit in Cypher but can be controlled explicitly with the xref:clauses/group-by.adoc[`GROUP BY` clause].
```
hvub: The added test helps. But sounds like grouping keys can be both implicit and explicit at the same time. They are either–or. Without GROUP BY, they are implicit (implicitly identified by their lack of aggregate function). With GROUP BY, they are explicit (explicitly identified/stated in the GROUP BY).

Here is a suggestion:

```suggestion
Grouping keys are either explicitly listed in the xref:clauses/group-by.adoc[`GROUP BY` clause] or implicit if no `GROUP BY` clause is given.
```

hvub (approved):
We start to get into fine tuning. Good work.

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: right, i think it's because of the way i queried the graph to make the nodes visible. will think of something and update :+1:

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
```diff
@@ -283,6 +283,11 @@ For example, GQL’s graph reference values `CURRENT_GRAPH` and `CURRENT_PROPERT
 | xref:clauses/order-by.adoc[`ORDER BY`]
 |
 
+| [[gq15]] GQ15
+| `GROUP BY` clause
+| xref:clauses/group-by.adoc[`GROUP BY`]
```
hvub, line 297: I have looked at the CIP now and tried the queries given there, specifically this one:

```
UNWIND [{a: {p: 1}, b: 20}, 
        {a: {p: 1}, b: 10}, 
        {a: {p: 2}, b: 10}] AS _row
WITH _row.a AS a, _row.b AS b
RETURN a.p AS p, {p: b} AS a 
  GROUP BY a, p
  ORDER BY a.p
```

It orders by the result by the `{p: b} AS a` column, this I think is and the CIP states as the GQL semantics. So I still think having no restriction here should be fine. We can double check with @JoelBergstrand when he is back and fix it later in case I made wrong conclusions. This does not have to hold up this PR.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+[[query-group-by]]
+= GROUP BY
+
+`GROUP BY` is a subclause that gives explicit control over the grouping keys of a `RETURN` or `WITH` clause.
+Cypher's implicit grouping stays the same.
+
+`GROUP BY` arranges identically structured data into groups based on one or more xref:functions/aggregating.adoc#grouping-keys[grouping keys].
+For this reason, it combines well with aggregate functions like xref:functions/aggregating.adoc#functions-count[count()] or xref:functions/aggregating.adoc#functions-avg[avg()] to perform calculations on each group of data.
+
+Cypher's `GROUP BY` functions similarly to SQL's `GROUP BY` and it is also xref:appendix/gql-conformance/supported-optional.adoc#gq15[GQL-compliant].
```
JPryce-Aklundh: We tend not to be in-text explicit about the GQL compliance-

```suggestion
Cypher's `GROUP BY` is similar to SQL's `GROUP BY`.
It arranges identically structured data into groups based on one or more xref:functions/aggregating.adoc#grouping-keys[grouping keys].
For this reason, it combines well with aggregate functions like xref:functions/aggregating.adoc#functions-count[count()] or xref:functions/aggregating.adoc#functions-avg[avg()] to perform calculations on each group of data.
```

modules/ROOT/pages/clauses/group-by.adoc
JPryce-Aklundh: Im wondering whether "implicit grouping" is clear to all readers? Maybe better to be a little bit more explicit?

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+
+
+== Grouping keys
```
JPryce-Aklundh:
```suggestion
[[grouping-keys]]
== Grouping keys
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+  (jessica)-[:ACTED_IN {role: 'Amy Loughren'}]->(goodnurse),
+  (james)-[:ACTED_IN {role: 'Bill Denbrough'}]->(it2),
+  (eddie)-[:ACTED_IN {role: 'CHarlie Cullen'}]->(goodnurse)
+----
+
+
+[[syntax]]
+== Syntax
+
+.Syntax for GROUP BY in a RETURN clause
```
JPryce-Aklundh:
```suggestion
.Syntax for `GROUP BY` in a `RETURN` clause
```

Same for below examples

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+| "Eddie Redmayne"   | "Drama"  | 2022 | 4.2           
+4+d|Rows: 7
+|===
+
+`GROUP BY ALL` uses all projected expressions as grouping keys.
+In this case, only the two drama movies from 2022 starring Jessica Chastain actually calculate an average from two values.
+=====
+
+
+== GROUP BY and ORDER BY
```
JPryce-Aklundh, line 208:
```suggestion
[[group-by-and-order-by]]
== GROUP BY and ORDER BY
```

Can we say something general about the interaction between these two? Just going straight into an example makes it unclear why we need a special section for this behavior.

Also, should this section be called out on the ORDER BY page?

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+
+The results are ordered by genre after being grouped.
+=====
+
+[NOTE]
+====
+Neo4j 2026.07 features changes to the visibility of expressions in `WHERE` and `ORDER BY` clauses when using `GROUP BY`:
+
+Variables named with `AS` in `RETURN` and `WITH` clauses are called _projected aliases_.
+Projected aliases make variables defined prior to `RETURN` or `WITH` in accessible, if these variables share the same names.
```
JPryce-Aklundh:
```suggestion
Projected aliases make variables defined prior to `RETURN` or `WITH` inaccessible if they share the same name.
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+====
+
+
+== Special cases
```
JPryce-Aklundh, line 250:
```suggestion
[[special-cases]]
== Special cases
```
Again, I think this needs a sentence or two stating what special cases there are instead of letting readers scroll through examples.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+| 2022 | 4.2    
+2+d|Rows: 7
+|===
+
+`title` is not returned, but still used as a grouping key.
+Without it, the result for 2019 with a rating of `3.6` would not be repeated.
+=====
+
+
+== Errors
```
JPryce-Aklundh, line 346:
```suggestion
[[errors]]
== Errors
```

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -27,6 +27,35 @@ For more information, see xref:queries/select-version.adoc[].
...
+[source, cypher, role="noheader"]
+----
+MATCH (a)-->(b)
+RETURN a.p AS a, COUNT(*) AS c 
+ORDER BY a.p
+----
+a| Deprecated the following:
+
+- subclause expressions which reference projection item expressions which are more complex than a variable or a property access on a variable.
+- cases with an ambiguity between an incoming variable and an alias in the subclause expression.
```
```suggestion
* Subclause expressions which reference projection item expressions which are more complex than a variable or a property access on a variable.
* Cases with an ambiguity between an incoming variable and an alias in the subclause expression.
```

Elsewhere you capitalised bulletpoints, but did not punctuate. (I think capitalisation + full-stops in bullet points are preferred?)

JPryce-Aklundh (commented):
Great work, @rsill-neo4j - just a few comments from me.

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+
+.Syntax for GROUP BY in a RETURN clause
+[source, syntax]
+----
+RETURN <return statement body>
+  GROUP BY { <grouping element> {, <grouping element> } … | () | ALL } 
+  <order by and page clause>
+----
+
+.Syntax for GROUP BY in a WITH clause
```
rsill-neo4j:
```suggestion
.Syntax for `GROUP BY` in a `WITH` clause
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+WITH <return statement body>
+  GROUP BY { <grouping element> {, <grouping element> } … | () | ALL } 
+  [<order by and page clause>]
+  [<where clause>]
+----
+
+[[examples]]
+== Basic examples
+
+.GROUP BY with a single grouping key
```
rsill-neo4j:
```suggestion
.`GROUP BY` with a single grouping key
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+        
+2+d|Rows: 3
+|===
+
+The movies are grouped by genre and the average rating for each genre is calculated.
+The four drama movies have an average rating of `4.225`.
+=====
+
+
+.GROUP BY with two grouping keys
```
rsill-neo4j:
```suggestion
.`GROUP BY` with two grouping keys
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+| "Eddie Redmayne"   | "Drama"  | 4.2                  
+        
+3+d|Rows: 5
+|===
+
+Grouping by actor and genre, the drama and horror movies have two entries each.
+=====
+
+
+.GROUP BY with no grouping keys
```
rsill-neo4j:
```suggestion
.`GROUP BY` with no grouping keys
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+2+d|Rows: 1
+|===
+
+When using `GROUP BY` with `()`, you explicitly state that there are no grouping keys.
+This means that there can be no direct projections of returned entities.
+However, aggregations still work like in the showcased query.
+=====
+
+
+.GROUP BY ALL
```
rsill-neo4j:
```suggestion
.`GROUP BY ALL`
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+|===
+
+`GROUP BY ALL` uses all projected expressions as grouping keys.
+In this case, only the two drama movies from 2022 starring Jessica Chastain actually calculate an average from two values.
+=====
+
+
+== GROUP BY and ORDER BY
+
+.GROUP BY and ORDER BY
```
rsill-neo4j:
```suggestion
.`GROUP BY` and `ORDER BY`
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+Neo4j 2026.07 features changes to the visibility of expressions in `WHERE` and `ORDER BY` clauses when using `GROUP BY`:
+
+Variables named with `AS` in `RETURN` and `WITH` clauses are called _projected aliases_.
+Projected aliases make variables defined prior to `RETURN` or `WITH` in accessible, if these variables share the same names.
+====
+
+
+== Special cases
+
+.GROUP BY without aggregation
```
rsill-neo4j:
```suggestion
.`GROUP BY` without aggregation
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+| 2022 | 4.2    
+2+d|Rows: 6
+|===
+
+In this query, `GROUP BY` is equivalent to using the xref:clauses/return.adoc#return-unique-results[`DISTINCT`] keyword.
+Two movies with a rating of `3.6` exist for the year 2019, but the entry is not repeated.
+=====
+
+
+.GROUP BY and aliasing
```
rsill-neo4j:
```suggestion
.`GROUP BY` and aliasing
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+| 2022 | 3.8    
+| 2022 | 4.2    
+2+d|Rows: 6
+|===
+
+You can group by aliased and unaliased return items (`year` is not aliased, `rating` is aliased).
+=====
+
+
+.GROUP BY an item that is not returned
```
rsill-neo4j:
```suggestion
.`GROUP BY` an item that is not returned
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+|===
+
+`title` is not returned, but still used as a grouping key.
+Without it, the result for 2019 with a rating of `3.6` would not be repeated.
+=====
+
+
+== Errors
+
+.Missing grouping element in GROUP BY
```
rsill-neo4j:
```suggestion
.Missing grouping element in `GROUP BY`
```

modules/ROOT/pages/clauses/group-by.adoc
```diff
@@ -0,0 +1,397 @@
...
+link:https://neo4j.com/docs/status-codes/current/errors/gql-errors/42001/[42001]: error: syntax error or access rule violation - invalid syntax
+|===
+
+`ORDER BY` cannot access `m` because it has been declared before the `RETURN` clause and the query contains a `GROUP BY` clause.
+
+For another example, see the last error described in xref:clauses/order-by.adoc#aggregation-distinct[Ordering aggregated or DISTINCT results].
+=====
+
+
+.GROUP BY with ambiguous references in grouping elements
```
rsill-neo4j:
```suggestion
.`GROUP BY` with ambiguous references in grouping elements
```

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: rephrased/added

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: added the callout and some text

modules/ROOT/pages/clauses/group-by.adoc
rsill-neo4j: added an unordered list of the special cases at the top of the section

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -27,6 +27,35 @@ For more information, see xref:queries/select-version.adoc[].
 [[cypher-deprecations-additions-removals-2026.07]]
 == Neo4j 2026.07
 
+=== Deprecation
```
```suggestion
=== Deprecations
```

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -85,6 +114,25 @@ It returns the number of keys in a `MAP`, the number of elements in a `LIST<ANY>
...
+
+a|
+label:new[]
+[source, cypher, role="noheader"]
+----
+MATCH (p:Person)-[:ACTED_IN]->(m:Movie)
+RETURN p.name AS actor_name, m.genre AS genre, avg(m.rating) AS averageRating
+GROUP BY actor_name, genre
+----
+| A new xref:clauses/group-by.adoc[GROUP BY] subclause used in conjunction with `WITH` and `RETURN` clauses has been added. It allows a user to explicitly state the grouping keys for the query.
```
```suggestion
| A new xref:clauses/group-by.adoc[`GROUP BY`] subclause used in conjunction with the `WITH` or `RETURN` clauses has been added. It allows a user to explicitly state the grouping keys for the query.
```

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -27,6 +27,35 @@ For more information, see xref:queries/select-version.adoc[].
...
+
+[source, cypher, role="noheader"]
+----
+MATCH (a)-->(b)
+RETURN a.p AS a, COUNT(*) AS c 
+ORDER BY a.p
+----
+a| Deprecated the following:
+
+* Subclause expressions which reference projection item expressions which are more complex than a variable or a property access on a variable.
```
```suggestion
* Subclause expressions which reference projection item expressions more complex than a variable or a property access on a variable.
```

modules/ROOT/pages/deprecations-additions-removals-compatibility.adoc
```diff
@@ -27,6 +27,35 @@ For more information, see xref:queries/select-version.adoc[].
...
+* Cases with an ambiguity between an incoming variable and an alias in the subclause expression.
```
```suggestion
* Cases where the subclause expression is ambiguous between an incoming variable and an alias.
```

JPryce-Aklundh (approved):

JPryce-Aklundh:
Also, @rsill-neo4j, remember to add cheat sheet tags 😃

neo4j-docops-agent:

modules/ROOT/pages/appendix/gql-conformance/supported-optional.adoc
JoelBergstrand: With GROUP BY the ORDER BY will use the alias `a` not the incoming `a`, e.g. it is correct that it is the value `b` that should be used for ordering - anything else is a bug.

PR #1609, opened 2026-07-15
SURF-1204: Explicitly state difference of + to || with toString

gem-neo4j:
We got a github issue showing confusion around why `+` allows non-string values and `||` does not, this PR makes this more explicit.

### DIFF
--- a/modules/ROOT/pages/expressions/string-operators.adoc
+++ b/modules/ROOT/pages/expressions/string-operators.adoc
@@ -7,8 +7,27 @@ Cypher contains two operators for the concatenation of `STRING` values:
 * `||`
 * `+`
 
-The two operators are functionally equivalent.
+The two operators are functionally equivalent on `STRING` values.
 However, `||` is xref:appendix/gql-conformance/index.adoc[GQL conformant], while `+` is not.
+
+When `+` is used to concatenate a `STRING` with a non-`STRING` value (excluding `LIST` values), the non-`STRING` value is implicitly converted to a `STRING` using xref:functions/string.adoc#functions-tostring[`toString()`].
+`||` does not perform this implicit conversion, so a `STRING` can only be concatenated with another `STRING` when using `||`; non-`STRING` values must first be explicitly converted using `toString()`.
+
+.Implicit `toString()` conversion with `+`
+[source, cypher]
+----
+RETURN 'The number is: ' + 42 AS result
+----
+
+.Result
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/expressions/string-operators.adoc
```diff
@@ -7,9 +7,28 @@ Cypher contains two operators for the concatenation of `STRING` values:
 * `||`
 * `+`
 
-The two operators are functionally equivalent.
+The two operators are functionally equivalent on `STRING` values.
 However, `||` is xref:appendix/gql-conformance/index.adoc[GQL conformant], while `+` is not.
 
+When `+` is used to concatenate a `STRING` with a non-`STRING` value (excluding `LIST` values), the non-`STRING` value is implicitly converted to a `STRING` using xref:functions/string.adoc#functions-tostring[`toString()`].
+`||` does not perform this implicit conversion, so a `STRING` can only be concatenated with another `STRING` when using `||`; non-`STRING` values must first be explicitly converted using `toString()`.
```
rsill-neo4j, line 14: on the surface level, this is probably because otherwise `||` would not be GQL conformant, but any idea about the reasoning in GQL? is there a paradigm of "purity of purpose"? like, let's not make the operator do obscure/implicit things (the conversion) when there's a dedicated function for it?

rsill-neo4j (approved):
also a general fix, right?

modules/ROOT/pages/expressions/string-operators.adoc
gem-neo4j: Yeah, it is for GQL conformance, hmm not sure, perhaps @hvub would be able to give a better response. I would assume it is to do with implicit vs explicit type conversion, I know GQL has a CAST keyword which we don't which may be a part of it.

hvub: Yeah, GQL ~does not have~ barely has overloaded operator, similar to SQL. Implicit, toString behavior GQL (and SQL) does not have at all. Such things the user has to write explicitly using CAST. The WG3 is a fan of explicity and not so much of implicity.

hvub (approved):
LGTM.

modules/ROOT/pages/expressions/string-operators.adoc
rsill-neo4j: thanks for elaborating :)

neo4j-docops-agent:

PR #1633, opened 2026-08-12
SURF-1322: Map Comprehension

gem-neo4j:
New feature: Map Comprehension :D e.g `RETURN {k: v IN MAP WHERE v IS NOT NULL | k: v}`


neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1633.surge.sh

Updated pages:
- [Additions, deprecations, removals, and compatibility](https://neo4j-docs-cypher-1633.surge.sh/cypher-manual/25/deprecations-additions-removals-compatibility)
- [Map expressions](https://neo4j-docs-cypher-1633.surge.sh/cypher-manual/25/expressions/map-expressions)

modules/ROOT/pages/expressions/map-expressions.adoc
```diff
@@ -460,3 +460,141 @@ RETURN keanu{.*} AS allKeanuProperties
...
+
+Map comprehension follows a similar syntax to xref:expressions/list-expressions.adoc#list-comprehension[list comprehension].
+
+.Syntax (map variant)
+[source, syntax]
+----
+{key: value IN map [WHERE predicate] | keyExpression : valueExpression}
+----
+
+The iteration step (`key: value IN map`) ensures that each key, value pair in a `map` is accessed one by one, while the `keyExpression` and `valueExpression` optionally applies a transformation to each `item`, creating new `MAP` of key, value pairs containing the modified elements.
```
hvub:
```suggestion
The iteration step (`key: value IN map`) ensures that each key-value pair in a `map` is accessed one by one, while the `keyExpression` and `valueExpression` optionally applies a transformation to each `item`, creating new `MAP` of key, value pairs containing the modified elements.
```

Btw: I would write key-value pair with an endash, i.e. key–value pair, cf. https://en.wikipedia.org/wiki/Dash#Relationships_and_connections. But I do not know if that is the docs policy.

modules/ROOT/pages/expressions/map-expressions.adoc
```diff
@@ -460,3 +460,141 @@ RETURN keanu{.*} AS allKeanuProperties
...
+
+
+.Syntax (list variant)
+[source, syntax]
+----
+{item IN list [WHERE predicate] | keyExpression : valueExpression}
+----
+
+The iteration step (`item IN list`) ensures that each element of a `list` is accessed one by one, while the `keyExpression` and `valueExpression` optionally applies a transformation to each `item`, creating new `MAP` of key, value pairs containing the modified elements.
```
hvub, line 488: Same here:

```suggestion
The iteration step (`item IN list`) ensures that each element of a `list` is accessed one by one, while the `keyExpression` and `valueExpression` apply a transformation to each `item`, creating new `MAP` of key-value pairs containing the modified elements.
```

modules/ROOT/pages/expressions/map-expressions.adoc
hvub, line 479:
```suggestion
The iteration step (`key: value IN map`) ensures that each key-value pair in a `map` is accessed one by one, while the `keyExpression` and `valueExpression` apply a transformation to each `item`, creating new `MAP` of key-value pairs containing the modified elements.
```
1) "key, value pair" is not a thing and inconsistent with the first paragraph that has "key-value pairs". Btw: I would write key-value pair with an endash, i.e. key–value pair, cf. https://en.wikipedia.org/wiki/Dash#Relationships_and_connections. But I do not know if that is the docs policy. 
2) The <map element projection> is not optional. The grammatical number of "applies" is also wrong.

modules/ROOT/pages/expressions/map-expressions.adoc
```diff
@@ -460,3 +460,141 @@ RETURN keanu{.*} AS allKeanuProperties
...
+
+
+.Considerations
+|===
+|Variables introduced inside a map comprehension (such as `key`, `value`, or `item`) shadow any outer variables with the same name and are not accessible outside the map comprehension.
+a|As `MAP` values cannot have duplicate keys, map comprehension follows a "last update wins" logic.
+If multiple elements result in the same key, the value from the last element processed will overwrite any previous value for that key.
```
hvub, line 495: Note that the order of the entries is not properly define for a map, so it is not properly define what the last update is. I guess that best approximation is that the entries of map `M` are processed in the same order in which the keys appear the list resulting from `keys(M)`, c.f. `RETURN keys({z: 1, a: 2}), keys({a: 1, z: 2})`. I have not verified that for map comprehension. Also it make an assumption about the implementation of maps — namely that the entries has some table order — which is not part of the language semantics. I am also not certain which order guarantees the underlying `MapValue` and all its implementations actual give. Looking at the `CombinedMapValue`, I am not sure this it has consistent iterator order for keySet and foreach. Maybe that does not surface here. But it illustrate the problem. So, it is a bit tricky what to write here. The safest contractually is to write "last update wins, but no guarantees for what the last entry is in a given map".

modules/ROOT/pages/expressions/map-expressions.adoc
```diff
@@ -460,3 +460,141 @@ RETURN keanu{.*} AS allKeanuProperties
...
+WITH {a: 1, b: 2, c: 3} AS map
+RETURN {k: v IN map | 'a' : v * 10} AS result
+----
+
+.Result
+[role="queryresult",options="header,footer",cols="1*<m"]
+|===
+| result
+
+| {a: 30}
```
hvub, line 546: This might also warrant a remark about the non-guarantee of the entry order.

hvub (approved):
Looks straightforward. Two sibling comments with a suggested fix. The only main issue I see is that "last update wins" suggesting a define order of the map entries, which Cypher does not really have. I think that needs to be marked with a caveat or so.

modules/ROOT/pages/expressions/map-expressions.adoc
```diff
@@ -460,3 +460,141 @@ RETURN keanu{.*} AS allKeanuProperties
...
+
+1+d|Rows: 1
+|===
+
```
JPryce-Aklundh, line 550: As per @hvub suggestion
 
```suggestion

The value in the result comes from whichever `MAP` entry is processed last (in this case, `{a:30}`).
That entry is not guaranteed, however, because Cypher does not define an order in which `MAP` entries are processed.
```

modules/ROOT/pages/expressions/map-expressions.adoc
JPryce-Aklundh: How about adding a final sentence to this part: 
```suggestion
a|As `MAP` values cannot have duplicate keys, map comprehension follows a "last update wins" logic.
If multiple elements result in the same key, the value from the last element processed will overwrite any previous value for that key.
However, the order in which elements are processed is not guaranteed, so which value is treated as "last" should not be relied upon.
```

modules/ROOT/pages/expressions/map-expressions.adoc
```diff
@@ -460,3 +460,141 @@ RETURN keanu{.*} AS allKeanuProperties
...
+----
+{item IN list [WHERE predicate] | keyExpression : valueExpression}
+----
+
+The iteration step (`item IN list`) ensures that each element of a `list` is accessed one by one, while the `keyExpression` and `valueExpression` optionally applies a transformation to each `item`, creating new `MAP` of key, value pairs containing the modified elements.
+
+
+.Considerations
+|===
+|Variables introduced inside a map comprehension (such as `key`, `value`, or `item`) shadow any outer variables with the same name and are not accessible outside the map comprehension.
```
JPryce-Aklundh, line 493: Same for remaining entries
```suggestion
| Variables introduced inside a map comprehension (such as `key`, `value`, or `item`) shadow any outer variables with the same name and are not accessible outside the map comprehension.
```

modules/ROOT/pages/expressions/map-expressions.adoc
```diff
@@ -460,3 +460,144 @@ RETURN keanu{.*} AS allKeanuProperties
...
+{item IN list [WHERE predicate] | keyExpression : valueExpression}
+----
+
+The iteration step (`item IN list`) ensures that each element of a `list` is accessed one by one, while the `keyExpression` and `valueExpression` apply a transformation to each `item`, creating new `MAP` of key-value pairs containing the modified elements.
+
+
+.Considerations
+|===
+| Variables introduced inside a map comprehension (such as `key`, `value`, or `item`) shadow any outer variables with the same name and are not accessible outside the map comprehension.
+a|As `MAP` values cannot have duplicate keys, map comprehension follows a "last update wins" logic.
```
gem-neo4j:
```suggestion
a| As `MAP` values cannot have duplicate keys, map comprehension follows a "last update wins" logic.
```

modules/ROOT/pages/expressions/map-expressions.adoc
```diff
@@ -460,3 +460,144 @@ RETURN keanu{.*} AS allKeanuProperties
...
+If multiple elements result in the same key, the value from the last element processed will overwrite any previous value for that key.
+However, the order in which elements are processed is not guaranteed, so which value is treated as "last" should not be relied upon.
+|Unlike list comprehension, the `keyExpression` and `valueExpression` (and the `\|` separator) are not optional.
```
gem-neo4j:
```suggestion
| Unlike list comprehension, the `keyExpression` and `valueExpression` (and the `\|` separator) are not optional.
```

modules/ROOT/pages/expressions/map-expressions.adoc
```diff
@@ -460,3 +460,144 @@ RETURN keanu{.*} AS allKeanuProperties
...
+|The `keyExpression` must evaluate to a `STRING` value or an error will be thrown.
```
gem-neo4j:
```suggestion
| The `keyExpression` must evaluate to a `STRING` value or an error will be thrown.
```

neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1633.surge.sh

Updated pages:
- [Additions, deprecations, removals, and compatibility](https://neo4j-docs-cypher-1633.surge.sh/cypher-manual/25/deprecations-additions-removals-compatibility)
- [Map expressions](https://neo4j-docs-cypher-1633.surge.sh/cypher-manual/25/expressions/map-expressions)

JPryce-Aklundh (approved):
Great! Thank you @gem-neo4j!

neo4j-docops-agent:
View the updated docs at https://neo4j-docs-cypher-1633.surge.sh

Updated pages:
- [Additions, deprecations, removals, and compatibility](https://neo4j-docs-cypher-1633.surge.sh/cypher-manual/25/deprecations-additions-removals-compatibility)
- [Map expressions](https://neo4j-docs-cypher-1633.surge.sh/cypher-manual/25/expressions/map-expressions)

neo4j-docops-agent:

PR #1634, opened 2026-08-13
DOCOPS-176 Promote all sections to docset level

recrwplay:
Use the new `page-tabs-promote-all` attribute to promote every section in _content-nav.adoc_ to the top level.

Instead of Cypher Manual listed under the Cypher tab, this has the effect of listing each section, using the same style.


neo4j-docops-agent:

